PR #3, opened 2019-01-16
updates to README

eckelcu:
The first commit contains some editorial suggestions. I also added a copyright statement to the end of the README. It is important that this appears somewhere. Alternatively, and actually a better option but a larger task, it to include the header info for the Apache license in all the source files. 

The second commit adds the DevNet Published badge to your repo. If you prefer to not have it, that is fine.


karmoham:
Thank you Charles!

PR #4, opened 2019-02-02
Parse show interface correctly on ASR1k

paneu:
On ASR1001-X, the built-in ports are of type BUILT-IN-2T+6X1GE which is not matched by the pattern for the type of the interface due to the fact, the + sign is not accounted for. Add this character to the regex so the parsing can be completed successfully.

Add a test case using exemplary data from a device as well.


siming85:
will address first thing monday

PR #5, opened 2019-02-03
Parse show version correctly on ISR4k

paneu:
On an ISR4451-X, the exact chassis identifier is called ISR4451-X/K9. We have to make sure everything of this string is matched by the regex, otherwise not all required keys of ShowVersionSchema will be parsed. Therefore, match for / characters as well, as this is missing at the moment.

Also, add a test case for show version using exemplary data from such a device.
As there are a lot of other tests for the different platform for the other show commands as well, please let me know whether I should contribute those as well.


lubnarasheed:
Merged into master

PR #6, opened 2019-02-06
Improve show spanning-tree summary parser for IOS XE

paneu:
This pull request contains changes, necessary to parse show spanning-tree summary on newer IOS XE platforms as there are certain features introduced that should be accounted for. Also, certain wording is different on the newer software release. Please review the new fields in the schema very carefully, as I do not know whether they exactly conform to the model. Bridge assurance seems to be part of the model as described in the PDF at https://pubhub.devnetcloud.com/media/pyats-packages/docs/genie/_models/stp.pdf, PVST simulation is not. Please feel free to give me any guidance on how to proceed here.
This PR also includes a minor fix for an edge case where only a single VLAN or MST instance is configured on a device.

Test coverage is added for the code as well.


lubnarasheed:
Thanks, great work! Looking forward to continue working together

PR #7, opened 2019-02-08
Minor fixes for OSPF parsing on IOS XE

paneu:
This MR contains two minor adjustments for behaviour of OSPF on IOS XE I noticed when running against a live environment. The commit messages should explain both of the changes in more detail.


lubnarasheed:
Thank you for your contributions!

PR #26, opened 2019-08-05
Corrections for ShowInterfacesSwitchport Class

jdrew82:
Corrected regex for VLAN names to allow special characters and save 
matched name in ShowInterfacesSwitchport class

Added handling for invisible characters at front of string for
ShowInterfacesSwitchport class

### DIFF
--- a/src/genie/libs/parser/iosxe/show_interface.py
+++ b/src/genie/libs/parser/iosxe/show_interface.py
@@ -1331,7 +1331,7 @@ class ShowInterfacesSwitchport(ShowInterfacesSwitchportSchema):
 
             # Access Mode VLAN: 1 (default)
             p8 =  re.compile(r'^Access +Mode +VLAN: +(?P<access_vlan>[\d\-]+)'
-                              '( *\((?P<dummy>\w+)\))?$')
+                              '( *\((?P<access_vlan_name>\w+)\))?$')
             m = p8.match(line)
             if m:
                 ret_dict[intf]['access_vlan'] = m.groupdict()['access_vlan']
@@ -1340,7 +1340,7 @@ class ShowInterfacesSwitchport(ShowInterfacesSwitchportSchema):
 
             # Trunking Native Mode VLAN: 1 (default)
             p9 =  re.compile(r'^Trunking +Native +Mode +VLAN: +(?P<native_vlan>[\d\-]+)'
-                              '( *\((?P<dummy>\w+)\))?$')
+                              '( *\((?P<native_vlan_name>\w+)\))?$')
             m = p9.match(line)
             if m:
                 if 'encapsulation' not in ret_dict[intf]:
--- a/src/genie/libs/parser/iosxe/show_interface.py
+++ b/src/genie/libs/parser/iosxe/show_interface.py
@@ -1187,9 +1187,11 @@ class ShowInterfacesSwitchportSchema(MetaParser):
                         Optional('administrative_encapsulation'): str,
                         Optional('operational_encapsulation'): str,
                         Optional('native_vlan'): str,
+                        Optional('native_vlan_name'): str,
                     },
                     Optional('negotiation_of_trunk'): bool,
                     Optional('access_vlan'): str,
+                    Optional('access_vlan_name'): str,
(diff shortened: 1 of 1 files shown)


jdrew82:
I found that your previous regex wasn't matching on the Access VLAN portion due to our use of special characters that aren't matched with \w. Using a .* allows it to match anything in the parenthesis. Also, for some reason there are invisible characters at the start of the string that are causing the line to not match. They appear to be '--More-- \x08\x08\x08\x08\x08\x08\x08\x08\x08        \x08\x08\x08\x08\x08\x08\x08\x08\x08' which breaks the parsing.

jeaubin:
@jdrew82  - Cool ! thanks! 

We will look into it and merge it next week.

src/genie/libs/parser/iosxe/show_interface.py
```diff
@@ -1333,6 +1335,7 @@ def cli(self, interface='', output=None):
             m = p8.match(line)
             if m:
                 ret_dict[intf]['access_vlan'] = m.groupdict()['access_vlan']
+                ret_dict[intf]['access_vlan_name'] = m.groupdict()['access_vlan_name']
```
lubnarasheed: It seems like the regex key name is `dummy`, please update the regex as well to have the key as `access_vlan_name`

https://github.com/CiscoTestAutomation/genieparser/blob/master/src/genie/libs/parser/iosxe/show_interface.py#L1332

src/genie/libs/parser/iosxe/show_interface.py
```diff
@@ -1343,6 +1346,7 @@ def cli(self, interface='', output=None):
                 if 'encapsulation' not in ret_dict[intf]:
                     ret_dict[intf]['encapsulation'] = {}
                 ret_dict[intf]['encapsulation']['native_vlan'] = m.groupdict()['native_vlan']
+                ret_dict[intf]['encapsulation']['native_vlan_name'] = m.groupdict()['native_vlan_name']
```
lubnarasheed: and the sam thing here about the key being `dummy` and not `native_vlan_name`
https://github.com/CiscoTestAutomation/genieparser/blob/master/src/genie/libs/parser/iosxe/show_interface.py#L1340

src/genie/libs/parser/iosxe/show_interface.py
jdrew82: Strange, thought I had already updated that. Must have been missed in my commit.

src/genie/libs/parser/iosxe/show_interface.py
jdrew82: Must have been missed in my commit. Will get corrected.

lubnarasheed: great! saw that the sample output line is already in

src/genie/libs/parser/iosxe/show_interface.py
lubnarasheed: great!

src/genie/libs/parser/iosxe/show_interface.py
jdrew82: Should be corrected now. Apologies about that.

src/genie/libs/parser/iosxe/show_interface.py
jdrew82: Should be corrected now. Apologies for that.

jdrew82:
I noticed that my other change for the regex somehow was missed in the commit so I've re-added it. The \w+ in the regex doesn't allow for use of : or _ which we use. Changing to .* allows for any characters in the ( ).

src/genie/libs/parser/iosxe/show_interface.py
lubnarasheed: Np, looks good now

src/genie/libs/parser/iosxe/show_interface.py
lubnarasheed: Np, looks good now, thanks!

lubnarasheed:
@jdrew82 thanks for enhancing this! It will be part of the next release

PR #36, opened 2019-08-23
Added IOS/IOS-XE parser for 'show vtp password'

jvoss:
Added parser for `show vtp password`

**Schema:**
```
{
  'vtp': {
     'configured': <class 'bool'>,
     Optional (str) 'password': <class 'str'>,
  }
}
```

This simple parser allows for checking whether a VTP password is configured and what the contents of the password string returned is (if configured).


jvoss:
Let me know if the dictionary key `vtp` is alright here. I was trying to remain consistent with `show vtp status` but considered `vtp_password` may be more appropriate?

PR #59, opened 2019-11-19
fixed IOS-XE reference in junos parser comment

oboehmer:
minor fix to change reference to IOS-XE in Junos parser ;-)


jeaubin:
Thanks! Will be part of the next release at the end of November (Last tuesday of every month is the release day)

PR #66, opened 2019-12-04
Fix `out referenced before assignment` in NX-OS `show access-lists`

supertylerc:
When the `output` variable is passed to the `cli` function for the `ShowAccessLists` class, an exception is raised because `out` is never assigned.  This fixes that to set `out` to the value of `output` when `output` is passed, which matches other classes/show commands.


src/genie/libs/parser/nxos/show_acl.py
```diff
@@ -130,6 +130,8 @@ def cli(self, acl="", output=None):
             else:
                 cmd = self.cli_command[0]
             out = self.device.execute(cmd)
+        else:
+            out = output
```
felipesasso: Good catch.

PR #67, opened 2019-12-09
Fix `cmd referenced before assignment` in IOSXE `show ip bgp summary`

supertylerc:
When the `output` variable is passed to the `cli` function for the `ShowIpBgpSummary` and `ShowIpBgpAllSummary` classes, an exception is raised because `cmd` is never assigned. This fixes that to set out to the value of `dmd` to an empty string (which matches the default of the called `super().cli()`) when output is passed.

### DIFF
--- a/src/genie/libs/parser/iosxe/show_bgp.py
+++ b/src/genie/libs/parser/iosxe/show_bgp.py
@@ -2297,7 +2297,6 @@ class ShowBgpAllSummary(ShowBgpSummarySuperParser, ShowBgpSummarySchema):
             # Execute command
             show_output = self.device.execute(cmd)
         else:
-            cmd = ''
             show_output = output
 
         # Call super
--- a/src/genie/libs/parser/iosxe/show_bgp.py
+++ b/src/genie/libs/parser/iosxe/show_bgp.py
@@ -2253,6 +2253,7 @@ class ShowBgpSummary(ShowBgpSummarySuperParser, ShowBgpSummarySchema):
             # Execute command
             show_output = self.device.execute(cmd)
         else:
+            cmd = ''
             show_output = output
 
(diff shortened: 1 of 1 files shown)


src/genie/libs/parser/iosxe/show_bgp.py
```diff
@@ -2253,6 +2253,7 @@ def cli(self, address_family='', vrf='', rd='', output=None):
             # Execute command
             show_output = self.device.execute(cmd)
         else:
+            cmd = ''
```
tahigash: thanks for pointing out this. we are using 'cmd' anyway, so better to do this before 'if' like below. please apply the same to other places.
```
    def cli(self, address_family='', vrf='', rd='', output=None):

        cmd = '' # <<<<<< HERE
        if output is None:
            # Build command
            if vrf:
                if address_family:
                    cmd = self.cli_command[0].format(address_family=address_family,
                                                 vrf=vrf)
            elif rd:
                if address_family:
                    cmd = self.cli_command[1].format(address_family=address_family,
                                                 rd=rd)
            elif address_family:
                cmd = self.cli_command[2].format(address_family=address_family)

            else:
                cmd = self.cli_command[3]
            # Execute command
            show_output = self.device.execute(cmd)
        else:
            show_output = output

supertylerc: Thanks for the feedback!  I believe this is addressed in ef3aa8f.

tahigash: @supertylerc thanks for the update!

tahigash:
codes look good

tahigash:
@supertylerc thanks for contribution again! merged.

PR #83, opened 2020-02-06
Show lldp issues

TopSecretAtWork:
I'm sorry I managed to fail to rebase my pull requests properly, so I closed the old ones and reopen them.

**NXOS:**

On some nxos devices we got results with no 'management_address_v6' information, like:

```
Chassis id: 547f.ee44.51e1 
Port id: mgmt:0 
Local Port id: mgmt0 
Port Description: mgmt0 
System Name: System1 
System Description: Cisco NX-OS n5000, Software (n5000-uk9), Version 7.3(2)N1(1), RELEASE SOFTWARE Copyright (c) 2002-2012, 2016-2017 by Cisco Systems, Inc. Compiled 5/12/2017 23:00:00 
Time remaining: 116 seconds 
System Capabilities: B 
Enabled Capabilities: B 
Management Address: 10.0.0.7 
Vlan ID: not advertised 
```

Because of that I propose to make the 'management_address_v6' field optional.

**IOS / IOSXE:**

On ios / iosxe devices we had issues parsing port id's containing a colon, like:

```
Port id: F8B7E2958F6F:P1
```

Because of that we changed the regex parsing the port id to the same pattern used for nxos devices, which can handle colons inside of the port id.

Also added tests, testing the cases:

NXOS
test_show_lldp_neighbors_detail_on_nxos

IOS / IOSXE
test_show_lldp_neighbors_detail_on_iosxe

### DIFF
--- a/src/genie/libs/parser/iosxe/show_lldp.py
+++ b/src/genie/libs/parser/iosxe/show_lldp.py
@@ -174,7 +174,7 @@ class ShowLldpEntry(ShowLldpEntrySchema):
         p1 = re.compile(r'^Local\s+Intf:\s+(?P<intf>[\w\/\.\-]+)$')
 
         # Port id: Gi1/0/4
-        p1_1 = re.compile(r'^Port\s+id:\s+(?P<port_id>\S+)$')
+        p1_1 = re.compile(r'^Port\s+id:\s+(?P<port_id>[\S\s]+)$')
 
         # Chassis id: 843d.c638.b980
         p2 = re.compile(r'^Chassis\s+id:\s+(?P<chassis_id>[\w\.]+)$')
--- a/src/genie/libs/parser/nxos/show_lldp.py
+++ b/src/genie/libs/parser/nxos/show_lldp.py
@@ -192,7 +192,7 @@ class ShowLldpNeighborsDetailSchema(MetaParser):
                                     }
                                 },
                                 'management_address_v4': str,
-                                'management_address_v6': str,
+                                Optional('management_address_v6'): str,
                                 Optional('system_capabilities'): str,
                                 Optional('enabled_capabilities'): str,
                                 'vlan_id': str
--- a/src/genie/libs/parser/iosxe/tests/test_show_lldp.py
+++ b/src/genie/libs/parser/iosxe/tests/test_show_lldp.py
@@ -1472,6 +1472,142 @@ class test_show_lldp_neighbor_detail(unittest.TestCase):
     'total_entries': 1,
 }
 
+    device_output_1 = {'execute.return_value': '''\
+            ------------------------------------------------
+            Local Intf: Gi4/0/38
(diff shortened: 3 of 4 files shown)


src/genie/libs/parser/iosxe/show_lldp.py
```diff
@@ -149,7 +149,7 @@ def cli(self, entry='',output=None):
 
         # initial regexp pattern
         p1 = re.compile(r'^Local +Intf: +(?P<intf>[\w\/\.\-]+)$')
-        p1_1 = re.compile(r'^Port +id: +(?P<port_id>[\w\/\.\-]+)$')
+        p1_1 = re.compile(r'^Port +id: +(?P<port_id>[\S ]+)$')
```
lubnarasheed: to match spaces we typically use ``\s+`` in the regex

src/genie/libs/parser/nxos/show_lldp.py
```diff
@@ -192,7 +192,7 @@ class ShowLldpNeighborsDetailSchema(MetaParser):
                                     }
                                 },
                                 'management_address_v4': str,
-                                'management_address_v6': str,
+                                Optional('management_address_v6'): str,
```
lubnarasheed: Is this new key being set by any regex in the parser? 
I don't see it in the parsed output example provided

src/genie/libs/parser/iosxe/show_lldp.py
TopSecretAtWork: Hi, thank you for your feedback.

I changed the regex in my last commit as you suggested (forgot to keep the changes after rebasing...)

I also fixed my test:

test_show_lldp_neighbors_detail_on_iosxev2

src/genie/libs/parser/nxos/show_lldp.py
TopSecretAtWork: It is not a new key, it already exists 
`p11 = re.compile(r'^Management +Address +IPV6: +(?P<mgmt_address_ipv6>.+)$')`

and it will be set here:

```python
# Management Address IPV6: not advertised
m = p11.match(line)
if m:
  group = m.groupdict()
  sub_dict.update({'management_address_v6': group['mgmt_address_ipv6']})

  continue
```

The `management_address_v6` field was mandatory before my change, but we got results from devices where no `management_address_v6` was set (as I mentioned in my opening post). Because of that I set the field to be optional.

src/genie/libs/parser/iosxe/show_lldp.py
lubnarasheed: thanks very much! looks good now

src/genie/libs/parser/nxos/show_lldp.py
lubnarasheed: Ah right, apologies I missed that

PR #90, opened 2020-02-24
Fix processing VRF names with dashes

miuvlad:
Many common configurations use dashes as part of the VRF name.
Because of the current regular expression used to match vrf names, these VRF names will be ignored.

The fix adds the full range of special characters as valid VRF names, as this is supported on IOS.

```
ip vrf !@#$%^&*()
ntp server vrf !@#$%^&*() 1.2.3.4
```


xiaoxinz-cisco:
Hello @miuvlad ,

You are absolutely right, the value of 'vrf' is various, there may be dashes, colons and other punctuations within the name.

Thank you for your contribution.

Sincerely,
Irene

PR #99, opened 2020-03-09
fix typo in NXOS show_bgp parser

ngrundler:


lubnarasheed:
Thank you for contributing!

PR #112, opened 2020-03-25
Fix NX-OS show platform parser

k01ek:
add support old n3k


KamyarZiabari:
Hi @k01ek ,

Thank you for the contribution. I looked at the code and everything seemed fine. Just one thing. As precaution. Can you please in this directory geneparser/tests’ dir run, this python -m unittest command and make sure that all the unittests are passing If so just take the screenshot and add to a comment. I will merge the code after that right away
Thank You

k01ek:
Hi @KamyarZiabari ! 
Thank you for replay. Of course i can run tests, but what python version should I use? And what about "automation"? With GitHub workflows you can do it easily.  I will create another PR for you.

PR #119, opened 2020-04-20
Adding more interface types to interface name coverter

noziwatele:
Added the following interface types and abbreviation variations:
'Ten': 'TenGigabitEthernet',
'Tw': 'TwoGigabitEthernet',
'Two': 'TwoGigabitEthernet',
'For': 'FortyGigabitEthernet',
'Hun': 'HundredGigE',

PR #126, opened 2020-04-24
Another interface type for the interface converter

noziwatele:
Hi all,
Sorry, just as soon as those last changes in #119 were merged I found another interface type that wasn't in the list. This is from `show cdp neighbors` in a Cat9500.

Twe = TwentyFiveGigE

I updated one of the test cases to cover this addition.

Test results:
```
Ran 3780 tests in 14.618s

OK
```

Please let me know if you need anything else.

Cheers!


src/genie/libs/parser/utils/common.py
```diff
@@ -208,6 +208,7 @@ def convert_intf_name(self, intf):
                    'Ten': 'TenGigabitEthernet',
                    'Tw': 'TwoGigabitEthernet',
                    'Two': 'TwoGigabitEthernet',
+                   'Twe': 'TwentyFiveGigE',
```
xiaoxinz-cisco, line 211: Hi @noziwatele , since other interfaces are all full names, could you please change it to `TwentyFiveGigabitEthernet` ?
Thanks!

noziwatele: Hi @xiaoxinz-cisco ,
I put `TwentyFiveGigE`, instead of the full name, because this is how the full version appears in the config and interface operational output in the Cat9500 that I have access to (C9500-48Y4C IOS-XE 16.9.4), so I thought that we would want to keep it the same as this to ensure command idempotency when genieparser output is used downstream.

Example config output:
```
c9500#sh run int Twe1/0/3
Building configuration...

Current configuration : 404 bytes
!
interface TwentyFiveGigE1/0/3
 description test
 switchport
 switchport trunk native vlan 900
 switchport trunk allowed vlan 1100,1150,1200,1250,1300,1400,1500,1600,1700
 switchport trunk allowed vlan add 1800,2100,2500
 switchport mode trunk
 switchport nonegotiate
 udld port aggressive
 channel-group 1 mode on
 spanning-tree portfast disable
 spanning-tree link-type point-to-point
end
```

Example "show interface" output:
```
c9500#sh int twe1/0/3
TwentyFiveGigE1/0/3 is up, line protocol is up (connected)
  Hardware is Twenty Five Gigabit Ethernet, address is 4ce1.755b.6503 (bia 4ce1.755b.6503)
  Description: test
  MTU 9154 bytes, BW 1000000 Kbit/sec, DLY 10 usec,
     reliability 255/255, txload 1/255, rxload 1/255
<...etc...>
```

Cheers!

xiaoxinz-cisco: Hi @noziwatele ,

Thanks for sharing the config and the output. They support that it's ok to put `TwentyFiveGigE` here.

Thanks,
Irene

PR #153, opened 2020-05-12
show power inline

TopSecretAtWork:
Hi,

we also wanted to be able to parse the output of `show power inline`. 
Therefore I changed the names from 'ShowPowerInlineInterface' to 'ShowPowerInline' (I copied that from 'ShowInterfaces').

I also added a new key to schema named 'watts' containing the available, used and remaining watts.
I changed the regex to also support white spaces in the device names.

Is there anything else I have to do because of renaming?

tests

### DIFF
--- a/src/genie/libs/parser/iosxe/show_power.py
+++ b/src/genie/libs/parser/iosxe/show_power.py
@@ -105,7 +105,8 @@ class ShowPowerInlineSchema(MetaParser):
 
 
 class ShowPowerInline(ShowPowerInlineSchema):
-    """Parser for show power inline """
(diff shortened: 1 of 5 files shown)


src/genie/libs/parser/iosxe/show_power.py
```diff
@@ -92,34 +92,62 @@ class ShowPowerInlineInterfaceSchema(MetaParser):
                 Optional('class'): str,
                 'max': float
             },
+        },
+        Optional('watts'): {
```
tahigash, line 96: added schema looks good

PR #167, opened 2020-06-15
Genieparser Issue - regex for VRF

medward2:
In our VRF naming convention we use AAA:123 format. VRFs containing the ‘:’ character seem to be missed by the genie parser.  As a quick test I changed the regex value in one of the pyATS parser files to this value - (?P<vrf_name>[A-Za-z0-9:]+) - Previously the pattern was - (?P<vrf_name>[A-Za-z0-9]+) 
 
This then corrected the problem, so looks like the RegEx in genie parser is not currently setup to allow non alpha numeric characters.

### DIFF
--- a/src/genie/libs/parser/nxos/show_cdp.py
+++ b/src/genie/libs/parser/nxos/show_cdp.py
@@ -29,7 +29,7 @@ class ShowCdpNeighborsSchema(MetaParser):
                     {'device_id': str,
                      'local_interface': str,
                      'hold_time': int,
-                     Optional('capability'): str,
+                     'capability': str,
                      'platform': str,
                      'port_id': str, }, }, },
     }
--- a/src/genie/libs/parser/nxos/show_cdp.py
+++ b/src/genie/libs/parser/nxos/show_cdp.py
@@ -27,9 +27,6 @@ class ShowCdpNeighborsSchema(MetaParser):
             {Optional('index'):
                 {Any():
                     {'device_id': str,
-                     'local_interface': str,
-                     'hold_time': int,
-                     'capability': str,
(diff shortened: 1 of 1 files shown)


xiaoxinz-cisco:
Hi @medward2 ,

Thanks for pointing out, can you create a new pull request from your branch please?
Here's a guide on creating a PR: https://help.github.com/en/github/collaborating-with-issues-and-pull-requests/creating-a-pull-request

In short, steps would be:
1. fork this pr
2. modify the code
- update the regex pattern 
- add sample unit test, for example: https://github.com/CiscoTestAutomation/genieparser/blob/master/src/genie/libs/parser/iosxe/tests/test_show_acl.py
3. update changelog/undistributed.rst
https://github.com/CiscoTestAutomation/genieparser/blob/master/changelog/undistributed.rst
4. push the code

I'll close this pull request because it doesn't include your fix.

Thanks,
Irene

medward2:
Hi,

Any chance you could provide some guidance on how I complete the unit tests? I see there is an existing test folder with unit test for test_show_interface.py.

Do I need to edit this file on GitHub or create a new version? I’m pretty new to Python programming so really a little lost here!

Regards
Mark

lubnarasheed:
Including @badalsoyantar to support

badalsoyantar:
Hi,

You made changes to ShowCdpNeighborsSchema. So you need to add unit test for ShowCdpNeighbors class.

Please see below link for other unit test for that class.
https://github.com/CiscoTestAutomation/genieparser/blob/master/src/genie/libs/parser/nxos/tests/test_show_cdp.py#L249

Thanks,
Badal

PR #180, opened 2020-07-08
Fix Helper address when there are more than 1 configured.

rara08:
Changed helper_address to array.
modified regex to capture 1+n helper ip's and append to array.

Note:  only tested on CAT9k_IOSXE (16.6.6) output.


rara08:
Fixed in beta release v20.6.1b0

PR #186, opened 2020-07-17
Iosxe/show stackwise virtual

itdependsnetworks:


changelog/2020/JULY.md
```diff
@@ -0,0 +1,11 @@
+| Module                  | Version       |
+| ------------------------|:-------------:|
+| ``genie.libs.parser``   |    20.7       |
+
+--------------------------------------------------------------------------------
+                                New
+--------------------------------------------------------------------------------
+
+* IOSXE
+    * Added ShowStackwiseVirtual for code issue:
```
tahigash, line 10: I think we can remove `for code issue`

src/genie/libs/parser/iosxe/tests/test_show_stackwise_virtual.py
```diff
@@ -0,0 +1,59 @@
...
+
+# ======================================
+# Unit test for 'show stackwise-virtual'
+# ======================================
+class test_show_stackwise_virtual(unittest.TestCase):
+    """Unit test for 'show stackwise-virtual'"""
+
+    maxDiff = None
+    empty_output = {"execute.return_value": ""}
+    golden_parsed_output1 = {
```
tahigash, line 16: based on below documentation, different `stackwise virtual link` is possible on same switch.
https://www.cisco.com/c/en/us/td/docs/switches/lan/catalyst3850/software/release/16-6/command_reference/b_166_3850_cr/b_166_3850_new_chapter_01011.html#wp1144390350

So, I would suggest to update the schema like below.
```
{
    "domain": 100,
    "enabled": True,
    "switches": {
        1: {
            "stackwise_virtual_link": {
                1: {
                    "ports": ['HundredGigE1/0/3', 'HundredGigE1/0/4']
(snip)

File: src/genie/libs/parser/iosxe/show_stackwise_virtual.py [hunk 3]
```diff
@@ -0,0 +1,110 @@
...
+        # Switch  Stackwise Virtual Link  Ports
+        # ------  ----------------------  ------
+        # 1       1                       TenGigabitEthernet1/0/47  
+        #                                 TenGigabitEthernet1/0/48  
+        # 2       1                       TenGigabitEthernet2/0/47  
+        #                                 TenGigabitEthernet2/0/48  
+
+        # Stackwise Virtual : Enabled
+        enabled_capture = "(?P<enabled>Enabled|Disabled)"
+        p_enabled = re.compile(f"Stackwise\s+Virtual\s+:\s+{enabled_capture}")
```
tahigash, line 57: We're afraid, but please change to `format()` instead of f-strings. pyATS still support python3.5 which doesn't support f-strings.

src/genie/libs/parser/iosxe/tests/test_show_stackwise_virtual.py
itdependsnetworks: Are lists preferred in this repo? I see a lot of `index` key's with values of integers.

tahigash: it's good question. if they are in the middle of hierarchy, we have those with `index` if needed. but in this case, they are leafs (last elements of tree/hierarchy). so we usually have them as `list` if no other items go under those ports.

itdependsnetworks: Can you refer me to an example? I do not know how the schema should look like in this scenario

tahigash: The schema would be like below.
```
    schema = {
        "domain": int,
        "enabled": bool,
        Optional("switches"): {
            int: {
                "stackwise_virtual_link": {
                    int: {
                        "ports": list
                    }
                }
            }
        }
    }

**Comment by itdependsnetworks** [comment]
@tahigash should be good to go

File: changelog/2020/JULY.md [hunk 4]
```diff
@@ -6,6 +6,120 @@
...
+    * Added ShowLDPInterfaceDetail
+        * show ldp interface {interface} detail
+    * Added PingMplsRsvp
+        * ping mpls rsvp {rspv}
+    * Added TracerouteNoResolve
+        * traceroute {ipaddress} no-resolve
+    * Added Ping
+        * ping {addr} ttl {ttl} count {count} wait {wait}
 * IOSXE
-    * Added ShowStackwiseVirtual for code issue:
```
tahigash: We released 20.7 yesterday. so JULY.md was updated...

At this point, can you move your changelog to below file? currently we are adding new ones to this file toward next release. (and I knew you already talked better idea with @jeaubin on other thread)
https://github.com/CiscoTestAutomation/genieparser/blob/master/changelog/undistributed.rst

PR #203, opened 2020-08-13
fix regex for extracting next hop uptime

nujo:
Fix regex for extracting next hop uptime from string which doesn't include outgoing interface. 
If next hop is like "[110/2] via 10.1.2.2, 06:46:59" the regex incorrecly assigns uptime to interface match group.  "?" needed after "," at the date match group to assign uptime to date match group


nujo:
Fix regex for extracting next hop uptime from string which doesn't include outgoing interface.
If next hop is like "[110/2] via 10.1.2.2, 06:46:59" the regex incorrecly assigns uptime to interface match group. "?" needed after "," at the date match group to assign uptime to date match group

PR #208, opened 2020-08-18
Iosxe tests

itdependsnetworks:
I still have not got to the tokens , and I skipper doing a few tests (as I either did not understand what they were doing, or haven't had the time to do yang yet) for now. I will add them, but these are massive changes already, don't need to add more to it.


itdependsnetworks:
Since many of the tests in IOS were pointing to IOSXE, i had to do them both at the same time.

itdependsnetworks:
Also, for the record, I regret doing iosxe 2nd, should have been the last one I did, but, I was into too deep :)

itdependsnetworks:
Can I get feedback on this? Would be good to merge this before any others, as I will constantly be running into merge conflicts otherwise.

jeaubin:
Looks good,  fix the conflict and you are good to go.

itdependsnetworks:
Will fix in a few minutes

itdependsnetworks:
ok, updated

Issue #231, opened 2020-09-09
NXOS 9.3(5) - C9332C "show spanning-tree detail" parser error


lukeod:
Hi,

Getting this error:

```
Missing keys: [['rapid_pvst', 'vlans', 1, 'interfaces', 'Port-channel1', 'timers'], ['rapid_pvst', 'vlans', 50, 'interfaces', 'Port-channel1', 'timers'], ['rapid_pvst', 'vlans', 51, 'interfaces', 'Port-channel1', 'timers'], << Ommited a ton of similar output >> ]]
  File "/opt/genie/poc/basic.py", line 88, in learn_stp
    abc = dev.parse('show spanning-tree detail')
  File "/opt/genie/poc/basic.py", line 354, in <module>
    aetest.main()

```
here's a sample of "show spanning-tree detail" on that host:

```
VLAN0001 is executing the rstp compatible Spanning Tree protocol
  Bridge Identifier has priority 0, sysid 1, address aaaa.bbbb.cccc
  Configured hello time 2, max age 20, forward delay 15
  We are the root of the spanning tree
  Topology change flag not set, detected flag not set
  Number of topology changes 0 last change occurred 162:37:09 ago
  Times:  hold 1, topology change 35, notification 2
          hello 2, max age 20, forward delay 15 
  Timers: hello 0, topology change 0, notification 0

 Port 4096 (port-channel1, vPC Peer-link) of VLAN0001 is root forwarding 
   Port path cost 100, Port priority 128, Port Identifier 128.4096
   Designated root has priority 1, address aaaa.bbbb.cccc
   Designated bridge has priority 0, address aaaa.bbbb.cccc
   Designated port id is 128.4096, designated path cost 0
   Timers: message age -38, forward delay 0, hold 0
   Number of transitions to forwarding state: 1
   The port type is network
   Link type is point-to-point by default
   BPDU: sent 292401, received 292410
```
This host is a Secondary host in a VPC. Perhaps negative timer values are to blame?

It's still occurring after i ran a "pyats version update".

```
[root@localhost poc]# pyats version check
You are currently running pyATS version: 20.8
Python: 3.6.8 [64bit]

  Package                      Version
  ---------------------------- -------
  genie                        20.8
  genie.libs.clean             20.8.1
  genie.libs.conf              20.8
  genie.libs.filetransferutils 20.8
  genie.libs.health            20.8
  genie.libs.ops               20.8
  genie.libs.parser            20.8
[...]

PR #235, opened 2020-09-14
NXOS show spanning-tree detail - handle negative timers

lukeod:
See https://github.com/CiscoTestAutomation/genieparser/issues/231


ahersi2-cisco:
Fix for this parser already made, beta will be released

Issue #239, opened 2020-09-14
NXOS 9.3(5) - C9332C "show interface" parser error


lukeod:
The parser for "show interface" is only picking up a subset of interfaces on my 9332C.

At a glance, it seems to be skipping interfaces that have this:
`Ethernet1/3 is down (XCVR not inserted)`

Example interface that is missing from the parser response:
```
Ethernet1/3 is down (XCVR not inserted)
admin state is down, Dedicated Interface
  Hardware: 10000/40000/100000 Ethernet, address: aaaa.bbbb.cccc (bia aaaa.bbbb.cccc)
  MTU 1500 bytes, BW 100000000 Kbit , DLY 10 usec
  reliability 255/255, txload 1/255, rxload 1/255
  Encapsulation ARPA, medium is broadcast
  Port mode is access
  auto-duplex, auto-speed
  Beacon is turned off
  Auto-Negotiation is turned on  FEC mode is Auto
  Input flow-control is off, output flow-control is off
  Auto-mdix is turned off
  Switchport monitor is off
  EtherType is 0x8100
  EEE (efficient-ethernet) : n/a
    admin fec state is auto, oper fec state is auto
  Last link flapped never
  Last clearing of "show interface" counters never
  0 interface resets
  Load-Interval #1: 30 seconds
    30 seconds input rate 0 bits/sec, 0 packets/sec
    30 seconds output rate 0 bits/sec, 0 packets/sec
    input rate 0 bps, 0 pps; output rate 0 bps, 0 pps
  Load-Interval #2: 5 minute (300 seconds)
    300 seconds input rate 0 bits/sec, 0 packets/sec
    300 seconds output rate 0 bits/sec, 0 packets/sec
    input rate 0 bps, 0 pps; output rate 0 bps, 0 pps
  RX
    0 unicast packets  0 multicast packets  0 broadcast packets
    0 input packets  0 bytes
    0 jumbo packets  0 storm suppression bytes
    0 runts  0 giants  0 CRC  0 no buffer
    0 input error  0 short frame  0 overrun   0 underrun  0 ignored
    0 watchdog  0 bad etype drop  0 bad proto drop  0 if down drop
    0 input with dribble  0 input discard
    0 Rx pause
  TX
    0 unicast packets  0 multicast packets  0 broadcast packets
    0 output packets  0 bytes
    0 jumbo packets
    0 output error  0 collision  0 deferred  0 late collision
[...]

PR #240, opened 2020-09-15
Fix issue with NXOS "show interface" parser - (XCVR not inserted)

lukeod:
See https://github.com/CiscoTestAutomation/genieparser/issues/239


danielgraziano:
Hi @lukeod can you please add a changelog and unittest? 

For the changelog, create a file called `changelog_feature_<uniquenumber>.rst` inside `https://github.com/CiscoTestAutomation/genieparser/tree/master/changelog/undistributed`. There is also a template located in the same folder for guidance.

For the unittest please add a new one or modify an existing to include 1 interface with this string. The tests for this parser are located here: https://github.com/CiscoTestAutomation/genieparser/blob/master/src/genie/libs/parser/nxos/tests/test_show_interface.py

Once the above is done I can review and merge. Thanks for contribution!

PR #241, opened 2020-09-15
update tests fix a few issues

itdependsnetworks:
Updated tests. There seems to be something wrong with the testing strategy as I had to fix several issues, and some tests do not seem like they were ever working. 

Also added an argparse on the pyats file to act as a "nosetest"

e.g. python `test_folder_parsing.py -o iosxe -c ShowVersion` works as one would expect


jeaubin:
Once conflicts merged (thanks as discussed)  let's merge it

itdependsnetworks:
Replace with https://github.com/CiscoTestAutomation/genieparser/pull/246 based on merge conflicts

PR #246, opened 2020-09-16
Tests take5

itdependsnetworks:


itdependsnetworks:
Updated tests. There seems to be something wrong with the testing strategy as I had to fix several issues, and some tests do not seem like they were ever working.

Also added an argparse on the pyats file to act as a "nosetest"

e.g. python test_folder_parsing.py -o iosxe -c ShowVersion works as one would expect

PR #256, opened 2020-09-19
ShowWlanSummary

itdependsnetworks:


src/genie/libs/parser/iosxe/tests/ShowWlanSummary/cli/equal/golden_output_expected.py
```diff
@@ -0,0 +1,31 @@
...
+            19: {
+                "profile_name": "internet_Global",
+                "ssid": "internet",
+                "wlan_status": "UP",
+                "status_security": "[open],MAC Filtering"
+            },
+            20: {
+                "profile_name": "lizzard-l_Global",
+                "ssid": "lizzard-legacy",
+                "wlan_status": "UP",
```
tahigash, line 26: just `status` is also OK.
```
"status": "UP",

File: src/genie/libs/parser/iosxe/tests/ShowWlanSummary/cli/equal/golden_output_expected.py [hunk 2]
```diff
@@ -0,0 +1,31 @@
...
+                "status_security": "[WPA2][802.1x][AES]"
```
tahigash, line 27: just `security`.
```
"security": "[WPA2][802.1x][AES]"

**Review by tahigash — CHANGES_REQUESTED** [review]
reviewed schema

File: src/genie/libs/parser/iosxe/show_wlan.py [hunk 3]
```diff
@@ -0,0 +1,106 @@
...
+            # ID   Profile Name                     SSID                             Status Security
+            elif wlan_info_header_capture.match(line):
+                wlan_info_header_capture_match = wlan_info_header_capture.match(line)
+                groups = wlan_info_header_capture_match.groupdict()
+                continue
+            # ----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
+            elif delimiter_capture.match(line):
+                delimiter_capture_match = delimiter_capture.match(line)
+                groups = delimiter_capture_match.groupdict()
+                continue
```
danielgraziano, line 86: This can be removed after removing the regex.

src/genie/libs/parser/iosxe/show_wlan.py
```diff
@@ -0,0 +1,106 @@
...
+        # 19   internet_Global                  internet                         UP     [open],MAC Filtering
+        # 20   lizzard-l_Global                lizzard-legacy                  UP     [WPA2][802.1x][AES]
+
+        # Number of WLANs: 4
+        wlan_count_capture = re.compile(r"^Number\s+of\s+WLANs:\s+(?P<wlan_count>\d+)$")
+        # ID   Profile Name                     SSID                             Status Security
+        wlan_info_header_capture = re.compile(r"^ID\s+Profile\s+Name\s+SSID\s+Status\s+Security$")
+        # ----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
+        delimiter_capture = re.compile(
+            r"^----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------$")
```
danielgraziano, line 61: This is not required, if a line doesn't match any regex then it will be skipped (which is the same as matching and doing nothing). Please remove.

danielgraziano (changes requested):

src/genie/libs/parser/iosxe/show_wlan.py
itdependsnetworks: I disagree with this design pattern, I actually think this leads to more issues over time. The positive affirmation of the project will cause issues in the long run. I took this from FAQ on ntc-templates (which I help maintain) 

```
Why is there a requirement to use Error in every template?

Ensuring that the textfsm template can account for every line is the only method to ensure that data was not accidentally missed. Take the following example. Initially we account for status to be:

Value STATUS (up|down)

Given the result of:

Interface                      Status         Protocol Description
Gi0/0/1                        admin down     down
Gi0/0/2                        up             up       ISP Connection
Gi0/0/3                        down           down
The output would miss the G0/0/1 interface, since the STATUS of admin down not known. If this was a low percentage use case, it can go undetected, and result in incorrect information being returned. Instead, by ensuring that we fail fast, an Error is raised and hopefully GitHub Issue is put in.
```

That being said, it is up to you, please respond either way.

salvapinyol: I fully agree with @itdependsnetworks. This parsing strategy makes the pyats parsing highly unreliable, especially for commands like show mac address-table, where some lines can be missed and is almost impossible to detect.

danielgraziano: @itdependsnetworks I see what you mean. As of current we don't raise an error if lines are not parsed, but it's a good idea that I've mentioned to the team. Keep the current implementation you have as it wont hurt.

src/genie/libs/parser/iosxe/tests/ShowWlanSummary/cli/equal/golden_output_expected.py
qduk: Updated

src/genie/libs/parser/iosxe/tests/ShowWlanSummary/cli/equal/golden_output_expected.py
qduk: Updated

ThomasJRyan (approved):

danielgraziano (approved):

src/genie/libs/parser/iosxe/tests/ShowWlanSummary/cli/equal/golden_output_expected.py
tahigash: confirmed

src/genie/libs/parser/iosxe/tests/ShowWlanSummary/cli/equal/golden_output_expected.py
tahigash: confirmed

tahigash (approved):

PR #262, opened 2020-09-22
Show redundancy regexissue iosxe

jmedina0911:


src/genie/libs/parser/iosxe/tests/ShowRedundancy/cli/equal/golden_output_cat4500_expected.py
```diff
@@ -0,0 +1,29 @@
+{
```
itdependsnetworks, line 1: Should say `expected_output = {` see https://github.com/CiscoTestAutomation/getting-started/blob/51d12c7e7ecc8ea1ce2a1f99e12a60903c70ab12/develop-guide/writeparser/writeparser.rst#test-your-new-parser for instructions.

jmedina0911: Perfect, It's already solved it, thanks

ahersi2-cisco (approved):

PR #288, opened 2020-10-07
Modified show class-map to account for specific name

qduk:
## Description
Modified show class-map to be able to accept a specific name passed to it.

## Motivation and Context
Allowed for show class-map {name} to be parsed

## Screenshots:
show_class

## Checklist:


src/genie/libs/parser/iosxe/show_class.py
```diff
@@ -16,8 +16,12 @@ class ShowClassMapSchema(MetaParser):
             str: {
               "match_criteria": str,
               "cm_id": int,
-              "match_policy": str,
+              Optional("match_policy"): str,
               Optional("description"): str,
+              Optional("match_access_group"): list,
+              Optional("match_dscp"): list,
+              Optional("match_cos"): list,
+              Optional("match_protocol_attribute"): list
```
tahigash, line 24: these needs to be under index like below:
```
"index": {
    int: {   # 1,2,3,...
        "match_policy": str,
        "match_access_group": list,
        "match_dscp": list,
        "match_cos": list,
        "match_protocol_attribute": list
```
because we could have multiple same match lines like below:
```
 Class Map match-all cos (id 1)
   Match cos  0 
   Match access-group  100 
   Match access-group  200 
   Match cos  1 
```
```
"index": {
    1: {
        "match_cos": [0],
    },
    2: {
        "match_access_group": [100],
    },
    3: {
        "match_access_group": [200],
    },
    4: {
        "match_cos": [1],
    }
```
or we can do like below as well. (having each match key under `match`. scheme will be deeper, but key name is cleaner.)
```
"index": {
    1: {
        "match": {
            "cos": [0],
        }
    },
    2: {
        "match": {
            "access-group": [100],
        }
    },
    3: {
        "match": {
            "access-group": [200],
        }
    },
    4: {
        "match": {
            "cos": [1],
        }
    }

File: src/genie/libs/parser/iosxe/show_class.py [hunk 2]
```diff
@@ -16,8 +16,12 @@ class ShowClassMapSchema(MetaParser):
             str: {
               "match_criteria": str,
               "cm_id": int,
-              "match_policy": str,
+              Optional("match_policy"): str,
```
tahigash, line 19: this needs to be under `index` as well.

tahigash (changes requested):

danielgraziano (approved):

src/genie/libs/parser/iosxe/show_class.py
qduk: This has been updated. I took the latter route because I liked the cleaner key names. Please let me know if it needs to be revised.

src/genie/libs/parser/iosxe/show_class.py
qduk: Added this to the index as well.

tahigash (approved):

PR #294, opened 2020-10-12
New Parsers for IOS-XR : show im dampening and show im dampening interface

gwoodwa1:
New parsers created for IOS-XR : _show im dampening_ and _show im dampening interface_

Tested using the new method (as per below). Note there is a warning related to another parser which appears to be flagged whenever I run unit testing.

(sdk) gary@gary-MS-7B17:~/sdk/genieparser/src/genie/libs/parser/iosxr/tests$ python -m unittest discover
................................................................................................../home/gary/sdk/genieparser/src/genie/libs/parser/iosxr/show_controllers.py:78: FutureWarning: Possible nested set at position 74
  p2 = re.compile(r'^mac\=(?P<mac>[A-Fa-f0-9:]+) +vlan=(?P<vlan>\d+)'
/home/gary/sdk/genieparser/src/genie/libs/parser/iosxr/show_controllers.py:78: FutureWarning: Possible nested set at position 176
  p2 = re.compile(r'^mac\=(?P<mac>[A-Fa-f0-9:]+) +vlan=(?P<vlan>\d+)'
......................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................
----------------------------------------------------------------------
Ran 600 tests in 1.270s

OK


src/genie/libs/parser/iosxr/tests/ShowImDampeningIntf/cli/equal/golden_output1_expected.py
```diff
@@ -0,0 +1,14 @@
+expected_output = {
+            'interface': {
+                'GigabitEthernet0/2/0/0': {
+                    'currently_suppressed': 'no',
```
tahigash, line 4: let's add 'interface_handle` here.
```
'GigabitEthernet0/2/0/0': {
    'interface_handle`: '0x080002c0',
```
corresponding output:
```
GigabitEthernet0/2/0/0 (0x080002c0)

File: src/genie/libs/parser/iosxr/tests/ShowImDampeningIntf/cli/equal/golden_output2_expected.py [hunk 2]
```diff
@@ -0,0 +1,31 @@
...
+                    'interface': {
+                        'TenGigE 0/1/0/0': {
+                            'currently_suppressed': 'yes',
+                            'dampening_status': 'enabled',
+                            'half_life': 1,
+                            'max_supress_time': 4,
+                            'penalty': 1625,
+                            'protocol': {
+                                'ipv4': {
+                                    'protocol_capsulation': 'ipv4',
```
tahigash, line 11: these ones has `protocol` level at line 9. so I believe we don't need prefix `procotol_`.
```
'protocol': {
    'ipv4': {
        'capsulation': 'ipv4',
        'penalty': 1615,
        'suppression': 'NO',
        'suppression_remaining_secs': 22,
        'underlying_state': 'Down',

File: src/genie/libs/parser/iosxr/tests/ShowImDampeningIntf/cli/equal/golden_output2_expected.py [hunk 3]
```diff
@@ -0,0 +1,31 @@
...
+                                    'protocol_penalty': 1615,
+                                    'protocol_suppression': 'no',
```
tahigash, line 13: let's take as is. no changing to lower case.
```
`suppression': 'NO'

File: src/genie/libs/parser/iosxr/tests/ShowImDampeningIntf/cli/equal/golden_output2_expected.py [hunk 4]
```diff
@@ -0,0 +1,31 @@
...
+                                    'protocol_suppression_timer': 22,
+                                    'protocol_underlying_state': 'down'
```
tahigash, line 15: same. take as is. `Down`

src/genie/libs/parser/iosxr/tests/ShowImDampeningIntf/cli/equal/golden_output1_expected.py
```diff
@@ -0,0 +1,14 @@
...
+                    'dampening_status': 'enabled',
+                    'half_life': 1,
+                    'max_supress_time': 4,
+                    'penalty': 0,
+                    'reuse': 750,
+                    'suppress': 2000,
+                    'underlying_state': 'up'
```
tahigash, line 11: take as is `Up`

src/genie/libs/parser/iosxr/tests/ShowImDampening/cli/equal/golden_output1_expected.py
```diff
@@ -0,0 +1,41 @@
...
+                                    'suppressed': 'no'
+                                }
+                            }
+                        },
+                        'GigabitEthernet0/0/0/1': {
+                            'protocol': {
+                                'not_present': {
+                                    'capsulation': 'not_present',
+                                    'penalty': 2389,
+                                    'suppressed': 'yes'
```
tahigash, line 17: take as is. `YES`

tahigash (changes requested):
thank you for your contribution. left comments about schema.

src/genie/libs/parser/iosxr/tests/ShowImDampening/cli/equal/golden_output1_expected.py
```diff
@@ -0,0 +1,41 @@
+expected_output = {
+                    'interface': {
+                        'GigabitEthernet0/0/0/0': {
+                            'protocol': {
+                                'not_present': {
```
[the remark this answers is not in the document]
gwoodwa1: Makes sense but I just wondered if that was a hardcoded key of "1:" ? It looks a bit like a list so I wondered if there was special way of coding of it or just treat it as normal set of dict keys. I think I am right in thinking we couldn't have more than one index (as we have the interface as a unique key) so hard coding it exactly like that should be okay?

src/genie/libs/parser/iosxr/tests/ShowImDampeningIntf/cli/equal/golden_output1_expected.py
gwoodwa1: Certainly, easy to add that into the schema.

src/genie/libs/parser/iosxr/tests/ShowImDampeningIntf/cli/equal/golden_output2_expected.py
gwoodwa1: Sure, I'll do that.. :)

src/genie/libs/parser/iosxr/tests/ShowImDampeningIntf/cli/equal/golden_output2_expected.py
gwoodwa1: Happy to do that, would you say the rule is leave all values "as is" and not to modify case or wording? I had another case example where a field just said "N" for No so I felt obliged to expand it to "No" so it was clear in terms of meaning. Just curious for anything I do in the future. Thank you.

src/genie/libs/parser/iosxr/tests/ShowImDampeningIntf/cli/equal/golden_output2_expected.py
gwoodwa1: Noted :)

src/genie/libs/parser/iosxr/tests/ShowImDampeningIntf/cli/equal/golden_output1_expected.py
gwoodwa1: Noted

src/genie/libs/parser/iosxr/tests/ShowImDampening/cli/equal/golden_output1_expected.py
gwoodwa1: Noted, all these comments will be corrected in the next update. Thanks.

src/genie/libs/parser/iosxr/show_dampening.py
```diff
@@ -0,0 +1,241 @@
...
+        typically contains 3 steps: exe
+        cuting, transforming, returning
+        """
+        if output is None:
+            out = self.device.execute(self.cli_command)
+        else:
+            out = output
+            
+        result_dict = {}
+        result_dict.setdefault('interface', {})
```
danielgraziano, line 52: I see where you are going with this. We need to move this inside the p1.match though. Reason being is we want the parser to return an empty dictionary if nothing parsed as that will raise a SchemaEmptyParserError.

src/genie/libs/parser/iosxr/show_dampening.py
```diff
@@ -0,0 +1,241 @@
...
+        
+        # GigabitEthernet0/0/0/1                                               2389 YES
+        # POS0/2/0/0                  <base>             ppp                      0 NO 
+        p1 = re.compile(r'^(?P<interface>.{1,26}) +(?P<prot>.{1,20}) +(?P<cap>.{1,10}) +(?P<pen>\d+) +(?P<sup>YES|NO)')
```
danielgraziano, line 56: I feel this regex may be a little risky (unless you are confident about the spacing) and it seems to catch a lot of whitespace:
image

`^(?P<interface>\S+) +(?P<prot>\S+)? +(?P<cap>\S+)? +(?P<pen>\d+) +(?P<sup>YES|NO)` Might be better:
image

src/genie/libs/parser/iosxr/show_dampening.py
```diff
@@ -0,0 +1,241 @@
...
+
+        for line in out.splitlines():
+            if line:
+                line = line.strip()
+            else:
+                continue
+            
+            m = p1.match(line)
```
danielgraziano, line 64: Please add an example on the line above for what this regex matches. Makes it easier to see at a glance

src/genie/libs/parser/iosxr/show_dampening.py
```diff
@@ -0,0 +1,241 @@
...
+            
+            if m:
+                group = m.groupdict()
+                interface=group['interface'].rstrip()
```
danielgraziano, line 68: rstrip() wont be needed with new regex

src/genie/libs/parser/iosxr/show_dampening.py
```diff
@@ -0,0 +1,241 @@
...
+                        group[empty]="not_present"
+                protocol=group['prot'].rstrip()
+                
+                if result_dict['interface'].get(interface)==None:
+                    result_dict['interface'][interface]={}
+                    result_dict['interface'][interface]['protocol']={}
+                result_dict['interface'][interface]['protocol'][protocol]={}
+                result_dict['interface'][interface]['protocol'][protocol].update({'capsulation': group['cap']})  
+                result_dict['interface'][interface]['protocol'][protocol].update({'penalty': int(group['pen'])})
+                result_dict['interface'][interface]['protocol'][protocol].update({'suppressed': group['sup'].lower()})
```
danielgraziano, line 82: We can improve this like below as setdefault() will insert data if it doesnt exist but it also returns a reference to a point within a dictionary:
``` python
protocol_dict = result_dict.setdefault('interface', {}).setdefault(interface, {}).setdefault('protocol', {})
protocol_dict.update({'capsulation': group['cap']})
protocol_dict.update({'penalty': int(group['pen'])})
protocol_dict.update({'suppressed': group['sup'].lower()})

File: src/genie/libs/parser/iosxr/show_dampening.py [hunk 13]
```diff
@@ -0,0 +1,241 @@
...
+                }
+        }
+    }
+    
+class ShowImDampeningIntf(ShowImDampeningIntfSchema):
+    """ Parser for show im dampening interface {interface} """
+    
+    cli_command = ['show im dampening interface {interface}']
+    
+    def cli(self, interface=None, output=None):
```
danielgraziano, line 117: lets make interface mandatory. Below we can then remove the `if interface:` line

src/genie/libs/parser/iosxr/show_dampening.py
```diff
@@ -0,0 +1,241 @@
...
+        """
+        if output is None:
+            if interface:
+                out = self.device.execute(self.cli_command[1].format(
+                                          interface=interface))
+        else:
+            out = output
+        
+        result_dict = {}
+        result_dict.setdefault('interface', {})
```
danielgraziano, line 131: Same as the other parser. We want an empty dictionary unless the parser actually parses lines

src/genie/libs/parser/iosxr/show_dampening.py
```diff
@@ -0,0 +1,241 @@
...
+        
+        # TenGigE 0/1/0/0 (0x01180020)
+        # GigabitEthernet0/2/0/0 (0x080002c0)
+        p1 = re.compile(r'^(?P<interface>[a-zA-Z].+[\/\d]+) \(')
```
danielgraziano, line 135: Please match the whole line (As @tahigash mentioned we will need to save the handle)

src/genie/libs/parser/iosxr/show_dampening.py
```diff
@@ -0,0 +1,241 @@
...
+        p4 = re.compile(r'^[a-z-A-Z\s]+state:\s+(?P<und_stat>\S+)')
+        
+        #  half-life: 1        reuse:             1000 
+        p5 = re.compile(r'half-life:\s+(?P<half_life>\d+) +reuse:\s+(?P<reuse>\d+)')
+        
+        # suppress:  1500     max-suppress-time: 4 
+        p6 = re.compile(r'suppress:\s+(?P<suppress>\d+) +max-suppress-time:\s(?P<max_suppress>\d+)')
+        
+        # ipv6           ipv6               1625  YES    42s  remaining        Down
+        p7 = re.compile(r'^(?P<prot>[a-z\d<>].{1,12}) +(?P<prot_cap>[a-z\d<>].{1,12})'
```
danielgraziano, line 154: Same as the other parser. Are you confident about the spacing in this line? Otherwise we should make it match more strictly
image

src/genie/libs/parser/iosxr/show_dampening.py
```diff
@@ -0,0 +1,241 @@
...
+        
+        for line in out.splitlines():
+    
+            if line:
+                line = line.strip()
+
+            else:
+                continue
+
+            m = p1.match(line)
```
danielgraziano, line 169: Please add comment above all regex to display line that can be matched

src/genie/libs/parser/iosxr/show_dampening.py
```diff
@@ -0,0 +1,241 @@
...
+            
+            if m:
+                group = m.groupdict()
+                interface=group['interface']
+                if result_dict['interface'].get(interface)==None:
```
danielgraziano, line 174: Same as other parser, we can improve the code flow

``` python
interface_dict = result_dict.setdefault('interface', {}).setdefault(interface, {})

# Then we can use the interface_dict reference in other matches below as well

**Review by danielgraziano — CHANGES_REQUESTED** [review]

File: src/genie/libs/parser/iosxr/tests/ShowImDampening/cli/equal/golden_output1_expected.py (hunk 7 above)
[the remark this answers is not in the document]
Comment (tahigash) [reply src/genie/libs/parser/iosxr/tests/ShowImDampening/cli/equal/golden_output1_expected.py:5]: the number `1` will be incremented if we see same interface as multiple. the index number will start with `1` and then `2`, `3` and so on.

We usually don't have a list in the middle of dictionary. (only leaf has list as value). So, that's why we are doing `index` thing for this case.

File: src/genie/libs/parser/iosxr/tests/ShowImDampeningIntf/cli/equal/golden_output2_expected.py (hunk 3 above)
Comment (tahigash) [reply src/genie/libs/parser/iosxr/tests/ShowImDampeningIntf/cli/equal/golden_output2_expected.py:13]: @gwoodwa1 thanks for the comment. actually, we did modification in some parsers in the past. but it created some confusions why it's changed when user is looking for exact output. so we now try to respect the output as is. I think `Y` or `N` is still understandable/guessable. if we had `E`(Enabled) or `D`(Disable), it's a bit hard to guess. In that kind of case, we should consider modification. but usually, will take values as is as much as possible.

Comment (gwoodwa1) [reply src/genie/libs/parser/iosxr/tests/ShowImDampeningIntf/cli/equal/golden_output2_expected.py:13]: Cool, thanks for the clarification. That guidance helps alot for the future.

File: src/genie/libs/parser/iosxr/show_dampening.py (hunk 8 above)
Comment (gwoodwa1) [reply src/genie/libs/parser/iosxr/show_dampening.py:52]: Makes sense, I've moved that inside p1.match now. Thanks

File: src/genie/libs/parser/iosxr/show_dampening.py (hunk 9 above)
Comment (gwoodwa1) [reply src/genie/libs/parser/iosxr/show_dampening.py:56]: Nice one Daniel, that suggestion is much better. I didn't know about the ? as an optional field. Thank you, I'll use that going forward.

File: src/genie/libs/parser/iosxr/show_dampening.py (hunk 12 above)
Comment (gwoodwa1) [reply src/genie/libs/parser/iosxr/show_dampening.py:82]: Hi Daniel, That works good. I didn't know from this that protocol_dict would also update result_dict as well. They look to be two separate dictionaries but clearly they are not. I am guessing this is because they share the same object in memory? However, whilst this has solved one problem it has created another in that empty fields e.g. capsulation has a value of "None"  which is busting my schema since it's not a string. I could just convert it to a string or change the schema to any() but is that okay? The same has happened in the protocol field as well where any empty fields are present. Data, below: {'interface': {'GigabitEthernet0/0/0/0': {'protocol': {None: {'capsulation': None, 'penalty': 629, 'suppressed': 'NO'}}}, 'GigabitEthernet0/0/0/1': {'protocol': {None: {'capsulation': None, 'penalty': 2389, 'suppressed': 'YES'}}}, 'POS0/2/0/0': {'protocol': {None: {'capsulation': None, 'penalty': 0, 'suppressed': 'NO'}, '<base>': {'capsulation': 'ppp', 'penalty': 0, 'suppressed': 'NO'}, 'ipv4': {'capsulation': 'ipcp', 'penalty': 0, 'suppressed': 'NO'}}}}}

File: src/genie/libs/parser/iosxr/show_dampening.py (hunk 15 above)
Comment (gwoodwa1) [reply src/genie/libs/parser/iosxr/show_dampening.py:135]: Got that one, thanks :+1:

File: src/genie/libs/parser/iosxr/show_dampening.py (hunk 12 above)
Comment (gwoodwa1) [reply src/genie/libs/parser/iosxr/show_dampening.py:82]: For now, I've made the schema for capsulation to be Any() and it works fine. If this is not right then please let me know. Cheers.

File: src/genie/libs/parser/iosxr/show_dampening.py (hunk 13 above)
Comment (gwoodwa1) [reply src/genie/libs/parser/iosxr/show_dampening.py:117]: Certainly, no problem have done that now in the updated version.

File: src/genie/libs/parser/iosxr/show_dampening.py (hunk 14 above)
Comment (gwoodwa1) [reply src/genie/libs/parser/iosxr/show_dampening.py:131]: Yeah, I've moved that inside the match p1 now instead. Cheers.

File: src/genie/libs/parser/iosxr/show_dampening.py (hunk 16 above)
Comment (gwoodwa1) [reply src/genie/libs/parser/iosxr/show_dampening.py:154]: Yeah, it's definitely looking better with your previous suggestion. Changed to the below:         
        p7 = re.compile(r'^(?P<prot>\S+)? +(?P<prot_cap>\S+)? +(?P<prot_pen>\d+) +(?P<prot_sup>YES|NO)'
                         ' +(?P<prot_sup_time>\d+). +remaining\s+(?P<prot_state>\S+)')

File: src/genie/libs/parser/iosxr/show_dampening.py (hunk 17 above)
Comment (gwoodwa1) [reply src/genie/libs/parser/iosxr/show_dampening.py:169]: Hi Daniel, Each re.compile line has a raw output line with a # comment to show what it is matching. Am I missing something? Could you point me at a specific example as I think I've done it already but maybe I am not seeing it? I've even shown two examples in some cases where the raw output varies e.g. underlying-state which has two ways of being outputted. underlying-state:  Up and Underlying state: Down. Thanks

File: src/genie/libs/parser/iosxr/show_dampening.py (hunk 18 above)
Comment (gwoodwa1) [reply src/genie/libs/parser/iosxr/show_dampening.py:174]: Changed in the newer version. Thanks.

File: src/genie/libs/parser/iosxr/tests/ShowImDampening/cli/equal/golden_output1_expected.py (hunk 7 above)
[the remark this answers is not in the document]
Comment (gwoodwa1) [reply src/genie/libs/parser/iosxr/tests/ShowImDampening/cli/equal/golden_output1_expected.py:5]: Since Daniel's changes, the empty key values now report None instead. I assume this would negate the need for using a index here? See conversation with Daniel higher up where those changes are proposed. If we need to convert from None to index then please let me know. Thanks.

File: src/genie/libs/parser/iosxr/show_dampening.py (hunk 12 above)
Comment (danielgraziano) [reply src/genie/libs/parser/iosxr/show_dampening.py:82]: Hi @gwoodwa1 for any optional items in the regex you can do:
``` python
if group['cap']:
  protocol_dict.update({'capsulation': group['cap']})
```
As we dont want to insert anything if it never existed

src/genie/libs/parser/iosxr/show_dampening.py
danielgraziano: Hi @gwoodwa1 we just like to have the same comment above both the re.compile and the .match(). Otherwise looking at this line I need to scroll up to the p1 definition to see what it matches. So do something like:

``` python
# TenGigE 0/1/0/0 (0x01180020)
# GigabitEthernet0/2/0/0 (0x080002c0)
p1 = re.compile(r'^(?P<interface>[a-zA-Z].+[\/\d]+) \(')

(snip)

# TenGigE 0/1/0/0 (0x01180020)
# GigabitEthernet0/2/0/0 (0x080002c0)
m = p1.match(line)

File: src/genie/libs/parser/iosxr/tests/ShowImDampening/cli/equal/golden_output1_expected.py (hunk 7 above)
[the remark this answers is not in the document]
Comment (danielgraziano) [reply src/genie/libs/parser/iosxr/tests/ShowImDampening/cli/equal/golden_output1_expected.py:5]: Hi @gwoodwa1, we still need the index instead of using protocol otherwise if there are two values with `None`, one would overwrite the other

**Comment by gwoodwa1** [comment]
I believe that I have I have addressed all the feedback and comments in the latest version. However, there was some significant amount of changes to the schema and code so please let me know if you think I missed anything. I've re-run the unit test successfully see below, Thank you for all your feedback as this is helping me improve the quality of these submissions.


(sdk) gary@gary-MS-7B17:~/sdk/genieparser/src/genie/libs/parser/iosxr/tests$ python -m unittest discover
................................................................................................../home/gary/sdk/genieparser/src/genie/libs/parser/iosxr/show_controllers.py:78: FutureWarning: Possible nested set at position 74
  p2 = re.compile(r'^mac\=(?P<mac>[A-Fa-f0-9:]+) +vlan=(?P<vlan>\d+)'
/home/gary/sdk/genieparser/src/genie/libs/parser/iosxr/show_controllers.py:78: FutureWarning: Possible nested set at position 176
  p2 = re.compile(r'^mac\=(?P<mac>[A-Fa-f0-9:]+) +vlan=(?P<vlan>\d+)'
......................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................
----------------------------------------------------------------------
Ran 600 tests in 1.252s

OK

File: src/genie/libs/parser/iosxr/tests/ShowImDampening/cli/equal/golden_output1_expected.py (hunk 7 above)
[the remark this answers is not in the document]
Comment (tahigash) [reply src/genie/libs/parser/iosxr/tests/ShowImDampening/cli/equal/golden_output1_expected.py]: confirmed the change

File: src/genie/libs/parser/iosxr/tests/ShowImDampeningIntf/cli/equal/golden_output1_expected.py (hunk 1 above)
Comment (tahigash) [reply src/genie/libs/parser/iosxr/tests/ShowImDampeningIntf/cli/equal/golden_output1_expected.py]: confirmed the change

File: src/genie/libs/parser/iosxr/tests/ShowImDampeningIntf/cli/equal/golden_output2_expected.py (hunk 2 above)
Comment (tahigash) [reply src/genie/libs/parser/iosxr/tests/ShowImDampeningIntf/cli/equal/golden_output2_expected.py]:
```
                            'index': {
                                1: {
                                    'capsulation': 'ipv6',
                                    'penalty': 1625,
                                    'protocol': 'ipv6',
                                    'suppression': 'YES',
                                    'suppression_timer': 42,
                                    'underlying_state': 'Down'
                                },
```
can I ask you change from `suppression_timer` to `suppression_remaining_sec`? better to have keyword `remaining` and also unit `secs`.

File: src/genie/libs/parser/iosxr/tests/ShowImDampeningIntf/cli/equal/golden_output2_expected.py (hunk 3 above)
Comment (tahigash) [reply src/genie/libs/parser/iosxr/tests/ShowImDampeningIntf/cli/equal/golden_output2_expected.py]: confirmed the change.

File: src/genie/libs/parser/iosxr/tests/ShowImDampeningIntf/cli/equal/golden_output1_expected.py (hunk 5 above)
Comment (tahigash) [reply src/genie/libs/parser/iosxr/tests/ShowImDampeningIntf/cli/equal/golden_output1_expected.py]: confirmed the change

File: src/genie/libs/parser/iosxr/tests/ShowImDampening/cli/equal/golden_output1_expected.py (hunk 6 above)
Comment (tahigash) [reply src/genie/libs/parser/iosxr/tests/ShowImDampening/cli/equal/golden_output1_expected.py]: confirmed the change

**Review by tahigash — CHANGES_REQUESTED** [review]
only 1 comment left from me.

File: src/genie/libs/parser/iosxr/tests/ShowImDampeningIntf/cli/equal/golden_output2_expected.py (hunk 2 above)
Comment (gwoodwa1) [reply src/genie/libs/parser/iosxr/tests/ShowImDampeningIntf/cli/equal/golden_output2_expected.py]: Sorry I missed that from the first round, I'll make that change.

Comment (gwoodwa1) [reply src/genie/libs/parser/iosxr/tests/ShowImDampeningIntf/cli/equal/golden_output2_expected.py]: I've made that change and uploaded an amended version for that. Thanks.

Comment (tahigash) [reply src/genie/libs/parser/iosxr/tests/ShowImDampeningIntf/cli/equal/golden_output2_expected.py]: confirmed the change

**Review by tahigash — APPROVED** [review]
schema looks good now.

File: src/genie/libs/parser/iosxr/show_dampening.py [hunk 19]
```diff
@@ -0,0 +1,257 @@
...
+# parser utils
+from genie.libs.parser.utils.common import Common
+
+# =============================================
+# Parser for 'show im dampening'
+# =============================================
+
+class ShowImDampeningSchema(MetaParser):
+    schema = {
+                'interface': {
```
danielgraziano, line 22: Please fix the indentation for this schema

src/genie/libs/parser/iosxr/show_dampening.py
```diff
@@ -0,0 +1,257 @@
...
+                    protocol_dict.update({'protocol': group['prot']})
+                continue
+   
+        return result_dict
+
+class ShowImDampeningIntfSchema(MetaParser):
+    schema = {
+        'interface': {
+            Any(): {
+            'dampening_status': str,
```
danielgraziano, line 95: Please fix the indentation for this schema

src/genie/libs/parser/iosxr/tests/ShowImDampeningIntf/cli/equal/golden_output1_expected.py
```diff
@@ -0,0 +1,15 @@
+expected_output = {
+                    'interface': {
```
danielgraziano, line 2: Please fix indentation

src/genie/libs/parser/iosxr/tests/ShowImDampeningIntf/cli/equal/golden_output2_expected.py
```diff
@@ -0,0 +1,34 @@
+expected_output = {
+                    'interface': {
```
danielgraziano, line 2: Please fix indentation

src/genie/libs/parser/iosxr/tests/ShowImDampeningIntf/cli/equal/golden_output3_expected.py
```diff
@@ -0,0 +1,8 @@
+
+expected_output = {
+                    'interface': {
```
danielgraziano, line 3: Please fix indentation

danielgraziano:
Hi @gwoodwa1, It looks great! Just fix the indentations that I left comments on and then we can go for merge! Thanks for the contribution!

gwoodwa1:
Fixed those indentation issues, I am going to leave auto PEP-8 on in future so those get automatically fixed. Hopefully, that is it. Thanks for your all your feedback and support.

src/genie/libs/parser/iosxr/tests/ShowImDampeningIntf/cli/equal/golden_output3_expected.py
gwoodwa1: Fixed

src/genie/libs/parser/iosxr/tests/ShowImDampeningIntf/cli/equal/golden_output2_expected.py
gwoodwa1: Fixed

src/genie/libs/parser/iosxr/tests/ShowImDampeningIntf/cli/equal/golden_output1_expected.py
gwoodwa1: Fixed

src/genie/libs/parser/iosxr/show_dampening.py
gwoodwa1: Fixed

src/genie/libs/parser/iosxr/show_dampening.py
gwoodwa1: Fixed

danielgraziano (approved):

PR #333, opened 2020-11-05
Addition of sros parser

paramountk:
## Description
Adds parser for show service sap-using command on sros

## Motivation and Context
allows data to be parsed

## Screenshots:
Screenshot 2020-11-05 at 22 56 13

## Checklist:


src/genie/libs/parser/sros/show_service_sap_using.py
```diff
@@ -0,0 +1,89 @@
...
+
+        result_dict = {}
+        # 2/1/6:10.*                      10         1     none    1     none   Up   Up
+        # 2/1/2:8.1                       181        1     none    1     none   Up   Down
+        # 2/1/2:3101.3101                 1121       3501  none    3501  none   Up   Down
+        # 2/1/6:*.*                    1121       1     none    1     none   Up   Up
+        # 2/1/6:3101.*                    1121       4503  none    4503  none   Up   Up
+        # lag-2:4000.*                    1122       1     none    1     none   Up   Up
+        # lag-1:400.1                    1221       1     none    1     none   Up   Up
+        p1 = re.compile(r'(?P<sap>(((\d+\/\d+\/\d+)|(lag-\d*))(:\d+|[*])*([.]\d+|[.*])*))'
```
ThomasJRyan, line 45: I admittedly don't know much about SROS. Will `(?P<sap>(((\d+\/\d+\/\d+)|(lag-\d*))(:\d+|[*])*([.]\d+|[.*])*))` cover all possibilities? Would a simpler `(?P<sap>\S+)` work better so we don't potentially have to go back and fix it should this have a different pattern?

src/genie/libs/parser/sros/show_service_sap_using.py
```diff
@@ -0,0 +1,89 @@
...
+                result_dict['sap'][sap].update({'oper_state': (group['oper'])})
+                continue
+
+            # Number of SAPs : 8
+
+            m = p2.match(line)
+            if m:
+                group = m.groupdict()
+                total = int(group['total'])
+                result_dict['total']=total
```
ThomasJRyan, line 87: Not a big thing, but this could be `result_dict['total'] = int(group['total'])`

src/genie/libs/parser/sros/show_service_sap_using.py
```diff
@@ -0,0 +1,89 @@
+from genie.metaparser import MetaParser
+from genie.metaparser.util.schemaengine import Any, Or, Optional
```
ThomasJRyan, line 2: Doesn't look like `Or` and `Optional` need to be imported

paramountk: ok, will remove.

src/genie/libs/parser/sros/show_service_sap_using.py
paramountk: Less is best, I will amend - thanks for this.

src/genie/libs/parser/sros/show_service_sap_using.py
paramountk: Hi, it will cover all standard ethernet type connections.
However, I prefer your simpler method as this should catch any differing format of sap as you suggest.

Many thanks for your feedback
Will amend.

paramountk: Hi, I need to add in support for some alternative hardware that I did not have access to.  Though SR-OS is generic, the alternative hardware will be missing one of the columns for this - so I am adding a secondary regex match and also making the variable optional in the schema.

ThomasJRyan: Sounds good. I don't see the changes yet but once they're in we can look at getting it merged

src/genie/libs/parser/sros/show_service_sap_using.py
```diff
@@ -0,0 +1,88 @@
...
+
+# ======================================================
+# Schema for 'show service sap-using'
+# ======================================================
+
+class ShowServiceSapUsingSchema(MetaParser):
+    """schema for show service sap-using"""
+    schema = {
+        'total': int,
+        'sap': {
```
tahigash, line 13: Just suggestion. based on output, `port_id` instead of `sap` is better?

tahigash (approved):
schema looks good. just 1 suggestion.

ThomasJRyan (approved):

tahigash:
thank you for your contribution!

PR #348, opened 2020-11-30
Update to New parser testing script(UPDATING)

ahersi2-cisco:
## Description
@itdependsnetworks - Thank you for creating this script as a new method to test our genieparsers. I made some additions to the script and would love you take a look at it when you have a moment. 

This PR includes changes to the ci_folder_parsing.py script that was developed as a new testing strategy for genie parser. Some of the additions that I made are mentioned below;

    
      
      I created a job file called 'folder_parsing_job.py' that would allow users to observe test results by running 'pyats logs view' in 
      the console after a run. The logs view makes it easier to navigate the results of the tests as opposed to scrolling through the 
      console to observe stdout, especially when you run the script to test all parsers. The tests can be executed by running the 
      command; 'pyats run job folder_parsing_job.py'

      I added the flag '-f' that also gets captured on the command line. Its purpose is to only display failed unittests to stdout. 
      Every time there is a change to a parser, or a new one is created, we have to make sure our new changes do not affect 
      existing parsers. We check this by running the script to test all parsers to see if any fail or error due to our new changes. 
      Since we have many parsers, if any parsers fail we have to scroll and search the terminal for the specific parser test that 
      fails. Adding the '-f' flag allows us to only display failed unittests to stdout.

      If you add '-v' to the end of the command line, it will display some more information that is normally hidden that includes 
      more detailed debugging information. I now included the results of the show command of the devices passed  
      parsedoutputs of that parser if that flag is included. For example, if you include the  '-f' in the command line, it will only 
      show failed outputs of the parsers. However if you also include the -v as well you will also see the outputs of the show 
      commands that passed.

      I am using pyATS's diff command to display the difference between the expected output and the developed output of the 
      parsers. It specifies where you exactly the differences are in parsed outputs.

## Motivation and Context
To allow for a better unittesting experience

## Impact (If any)
No negative impact

## Checklist:

- [X] I have updated the changelog.
- [X] All new and existing tests passed.


tests/ci_folder_parsing.py
```diff
@@ -385,4 +334,281 @@ def test_empty(self, steps, local_class, operating_system, token=None):
...
+    _number = args.number
+
+    if _number and (not _class or not _number):
+        sys.exit("Unittest number provided but missing supporting arguments:"
+                "\n* '-c' or '--class_name' for the parser class"
+                "\n* '-o' or '--operating_system' for operating system")
+
+    # This is the list of Classes that currently have no testing. It was found during the process
+    # of converting to folder based testing strategy
+    CLASS_SKIP = {
```
itdependsnetworks, line 375: This seems to be duplicated,

itdependsnetworks:
Looks great!! I left the one comment, I think constants should likely stay at the top in global. I didn't go deep into the code, I will trust the logic makes sense.

ahersi2-cisco:
Sounds great, thank you!

tests/ci_folder_parsing.py
```diff
@@ -271,9 +170,9 @@ def check_os_folder(self, steps, operating_system, number=_number):
                 # Each "globals()" is checked to see if it has a cli attribute, if so, assumed to be a parser. The _osxe, is
                 # since the ios module often refers to the iosxe parser, leveraging this naming convention.
                 if hasattr(local_class, "cli") and not name.endswith("_iosxe"):
-                    #if name == 'SomeClassName':
+                    # if name == 'SomeClassName':
                     #    start = True
-                    #if not start:
+                    # if not start:
                     #    continue
```
ThomasJRyan, line 176: Old code or comments? If it's code then probably best to remove it to clean things up, if it's comments then all is good

xiaoxinz-cisco (approved):

tests/ci_folder_parsing.py
```diff
@@ -385,4 +334,280 @@ def test_empty(self, steps, local_class, operating_system, token=None):
...
+    _number = args.number
+
+    if _number and (not _class or not _number):
+        sys.exit("Unittest number provided but missing supporting arguments:"
+                "\n* '-c' or '--class_name' for the parser class"
+                "\n* '-o' or '--operating_system' for operating system")
+
+    # This is the list of Classes that currently have no testing. It was found during the process
+    # of converting to folder based testing strategy
+    CLASS_SKIP = {
```
ThomasJRyan, line 375: What if we were to put this information into a separate yaml file instead of keeping it as a hard coded dictionary? Would make it easier to and and remove entries as unittests are developed I'd think. Same think with `EMPTY_SKIP`

tests/folder_parsing_job.py
```diff
@@ -0,0 +1,50 @@
+import os
+import sys
+import argparse
+from ats.datastructures.logic import And, Not, Or
+
+from genie.harness.main import gRun
```
ThomasJRyan, line 6: Probably don't need to import `gRun` here

ThomasJRyan (approved):

tests/ci_folder_parsing.py
ahersi2-cisco: This is a comment

tests/ci_folder_parsing.py
ahersi2-cisco: That makes sense, either yaml or txt, I'll add that

itdependsnetworks: it would be good to verify this list again, I think some were pending a PR as an example, and there may be tests already.

PR #382, opened 2021-01-23
Bugfix/version as a tuple nxos

rohit04saluja:
## Description
```system_verions``` shows the complete version of the system along with build in a long string format. But the software version is typically a tuple of integers of the form ```10.1.1```.
To achieve this following steps are implemented:
1. Split ```system_verion``` with delimiter ```' '```. Pick 0th element from the split.
2. Split the new value using regular expression split with delimiters, ```','```, ```'('``` and ```')'```.
3. Remove empty elements from the new list.
4. Convert each element of the list to an integer value using ```int()```. If the element can't be converted then leave it as a string.
5. Convert the modified list into a tuple to keep it unmutable.

## Motivation and Context
The change provides the version in a tuple of int and str. This helps in version checks for features where int operators can be used. This is particularly useful in the latest version of nxos where the tuple generated will be a pure int tuple.

## Impact (If any)
None

## Screenshots:
```
$ python -m unittest nxos.test_show_platform.test_show_version -v 
test_empty (nxos.test_show_platform.test_show_version) ... ok
test_golden (nxos.test_show_platform.test_show_version) ... ok
test_golden2 (nxos.test_show_platform.test_show_version) ... ok
test_golden3 (nxos.test_show_platform.test_show_version) ... ok
test_golden4 (nxos.test_show_platform.test_show_version) ... ok

----------------------------------------------------------------------
Ran 5 tests in 0.012s

OK
```

## Checklist:
- [*] I have updated the changelog.
- [*] I have updated the documentation (If applicable).
- [*] I have added tests to cover my changes (If applicable).
- [*] All new and existing tests passed.
- [*] All new code passed compilation.

### DIFF
--- a/src/genie/libs/parser/nxos/tests/test_show_platform.py
+++ b/src/genie/libs/parser/nxos/tests/test_show_platform.py
@@ -49,6 +49,7 @@ class test_show_version(unittest.TestCase):
                                  'kickstart_compile_time': '4/30/2017 23:00:00 [04/15/2017 ''04:34:05]',
                                  'kickstart_image_file': 'slot0:///n7000-s2-kickstart.10.81.0.129.gbin',
                                  'system_version': '8.1(1) [build 8.1(0.129)] [gdb]',
+                                 'version': (8, 1, 1),
                                  'system_compile_time': '4/30/2017 23:00:00 [04/15/2017 ''06:43:41]',
                                  'system_image_file': 'slot0:///n7000-s2-dk10.34.1.0.129.gbin'}
                               }
@@ -75,6 +76,7 @@ class test_show_version(unittest.TestCase):
                                  'seconds': 37},
                               'software':
                                 {'system_version': '7.0(3)I5(2) [build 7.0(3)I5(1.145)]',
+                                 'version': (7, 0, 3, 'I5', 2),
                                  'system_compile_time': '1/4/2017 20:00:00 [01/04/2017 21:47:15]',
                                  'system_image_file': 'bootflash:///ISSUCleanGolden.system.gbin'}
                               }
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Comment by rohit04saluja** [comment]
This resolved #380

File: src/genie/libs/parser/nxos/tests/test_show_platform.py [hunk 1]
```diff
@@ -49,6 +49,7 @@ class test_show_version(unittest.TestCase):
                                  'kickstart_compile_time': '4/30/2017 23:00:00 [04/15/2017 ''04:34:05]',
                                  'kickstart_image_file': 'slot0:///n7000-s2-kickstart.10.81.0.129.gbin',
                                  'system_version': '8.1(1) [build 8.1(0.129)] [gdb]',
+                                 'version': (8, 1, 1),
```
tahigash, line 52: main purpose of parser is, to get the value as is without modification. if you need something like this, maybe having API for it would be best? like API `get_version(format='tuple')`

rohit04saluja: This should also do the trick I believe. I might cause confusion between ```['platform']['version']``` and ```['platform']['software']['version']```

tahigash: so you will open PR for API instead?

rohit04saluja: The API will return the version for ```['platform']['version']``` or ```['platform']['software']['version']``` ?

rohit04saluja: @tahigash please check #386

tahigash: replied on the PR.
https://github.com/CiscoTestAutomation/genieparser/pull/386#issuecomment-770249130

tahigash: @rohit04saluja maybe we can close this PR? because you are working on API in genielibs. not in genieparser.

PR #402, opened 2021-02-26
Update README.md

LukasMcClelland:
Added `python ci_folder_parsing.py` to checklist

## Description

Added  ci_folder_parsing.py to readme

## Motivation and Context

New and old types of tests should be run

## Impact (If any)

none

## Screenshots:

N/A

## Checklist:

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,8 +51,7 @@ YouTube Video: <How to write a Genie parser for Cisco!> https://youtube.com/watc
 Once you create a new parser, don't forget to check 1, 2, and 3;
 if you only update the parser class without modifying/creating the schema, please check 2 and 3. 
 - [ ] 1. `make json`
-- [ ] 2. cd tests; and execute `python -m unittest -v` and `python ci_folder_parsing.py`
-- [ ] 3. create changelog for your pull request.
+- [ ] 2. create changelog for your pull request.
 
 # How to write 'changelog':
 1. A few examples are added into changelog/undistributed/template.rst:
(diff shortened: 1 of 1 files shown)


renatoalmeidaoliveira:
Why not:
`pyats run job folder_parsing_job.py -v -o <os> -c <class>`
That way the test runs only over the modified class and os, and prints the expected and the executed results?

tahigash:
> Why not:
> `pyats run job folder_parsing_job.py -v -o <os> -c <class>`
> That way the test runs only over the modified class and os, and prints the expected and the executed results?

@renatoalmeidaoliveira you're right. to test only added parser, that's the best syntax. but here we are trying to ask to test everything **just in case**.

@LukasMcClelland but technically, we are doing both `python -m unittest -v` and `python ci_folder_parsing.py` via GitHub Actions. so we might not need to ask contributor to do this anymore.

README.md
```diff
@@ -51,8 +51,7 @@ YouTube Video: <How to write a Genie parser for Cisco!> https://youtube.com/watc
 Once you create a new parser, don't forget to check 1, 2, and 3;
 if you only update the parser class without modifying/creating the schema, please check 2 and 3.
```
tahigash, line 52: need to update here, it says 1,2 and 3... and also please check other location as well to adjust to this change.

Let's add 3 like:
```
- [ ] 3. make sure GitHub Actions checks passed.

LukasMcClelland: Fixed

tahigash (approved):

tahigash:
@LukasMcClelland please get 2nd approval on this.

jeaubin (approved):

PR #496, opened 2021-07-26
Add new parser logic and commands to nxos ShowRunInterface

zhangineer:
## Description

Added support for 2 new commands to gather interface configurations on NX-OS devices. Current genie parser supports supports only single interface

command#1 : show running-config interface (This command applies to only NX-OS and allows all interface configurations to be gathered)
command#2: show running-config | section ^interface. (This command produce similar result as the above, however, this command can be run on both NX-OS and IOS (to be added), this will also allow some flexibility for users that they can choose either commands to gather interface)

Additionally, added 4 new parser logic under the same parser class.

All changes were made to the same parser class.

## Motivation and Context

This change will allow users to pull the entire interface configurations of a device. This is useful for assessment efforts, audits, configuration diff...etc.

## Impact (If any)

No negative impact

## Screenshots:

image

## Checklist:


src/genie/libs/parser/nxos/show_interface.py
```diff
@@ -3092,28 +3099,48 @@ class ShowRunningConfigInterfaceSchema(MetaParser):
...
+                   'show running-config | section ^interface',
+                   'show running-config interface']
+
+    def cli(self, interface=None, output=None):
+        # Determine command
         if output is None:
-            out = self.device.execute(self.cli_command.format(interface=interface))
+            if interface:
+                cmd = self.cli_command[0].format(interface=interface)
+            elif 'section ^interface' in self.cli_command:
```
ThomasJRyan, line 3132: This will never return `True`. I took a quick look and I'm not certain where (or if) the command is passed so that you could check this. If the produce the same results you can simply use one of the other in an `else`

src/genie/libs/parser/nxos/show_interface.py
```diff
@@ -3092,28 +3099,48 @@ class ShowRunningConfigInterfaceSchema(MetaParser):
...
+                cmd = self.cli_command[1]
+            else:
+                cmd = self.cli_command[2]
+            # Execute command
+            output = self.device.execute(cmd)
+            out = output
```
ThomasJRyan, line 3138:
```suggestion
            out = self.device.execute(cmd)
```

src/genie/libs/parser/nxos/show_interface.py
```diff
@@ -3276,14 +3303,45 @@ def cli(self, interface, output=None):
                 interface_dict.update({'description': m.groupdict()['description']})
                 continue
 
-            # vpc
+            # vpc ID, for port-channels only
             p15 = re.compile(r'^vpc +(?P<vpc>\S+)$')
```
ThomasJRyan, line 3307: I know you didn't make this parser initially, but if you could move all these `re.compile`s to before the loop that would be greatly appreciated

src/genie/libs/parser/nxos/tests/test_show_interface.py
```diff
@@ -1424,6 +1424,85 @@ class TestShowRunInterface(unittest.TestCase):
         },
     }
 
+    golden_output_5 = {'execute.return_value': '''
```
ThomasJRyan, line 1427: We no longer use this form of unittests. You can see an example of how we now do them here: https://github.com/CiscoTestAutomation/genieparser/tree/master/src/genie/libs/parser/nxos/tests/ShowRunningConfigNvOverlay/cli

ThomasJRyan (changes requested):
Overall an excellent contribution and we thank you for you. Mostly it's just the unittest that needs fixing. Everything else would just be nice to haves

zhangineer:
Thanks for the feedback @ThomasJRyan , I updated the code and tests based on your comments. Unit testing screenshots have been updated as well

ThomasJRyan (approved):

ThomasJRyan:
Thank you very much for making the changes. It is very much appreciated

LukasMcClelland:
Good PR, approved. Please resolve the merge conflict

ThomasJRyan:
@zhangineer please fix merge conflicts and we can get this merged

zhangineer:
@ThomasJRyan I tried merging this locally on my machine but did not see any conflict....any ideas ? I first fetched upstream master and then merged to my local repo

ThomasJRyan:
Thank you for your contribution @zhangineer

PR #498, opened 2021-07-29
Fix ShowBgpInstanceNeighborsDetail to match 'no-prepend' 'replace-as' and 'dual-as'

ancker010:
## Description
The ShowBgpInstanceNeighborsDetail p3 parser wasn't matching for the 'no-prepend' 'replace-as' and 'dual-as' that may be present on the Remote AS/Local AS line. Further, I've added lines to make sure state, either True or False for each flag is added to the output dictionary.

## Motivation and Context
When any of these flags are configured, the parser is unable to gather the Remote AS, Local AS, or any of the flags. And thus doesn't include them in the output dictionary.

## Impact (If any)
N/A

## Screenshots:
Screen Shot 2021-07-29 at 3 24 15 PM
More:
Screen Shot 2021-07-29 at 3 40 20 PM

## Checklist:


ancker010:
I looked through why the Run Test failed....it seems to be related to some iosxe thing that has nothing to do with my commits.
As you can see, the iosxr unittests passed.

Screen Shot 2021-07-29 at 3 57 35 PM

src/genie/libs/parser/iosxr/show_bgp.py
```diff
@@ -2673,10 +2671,18 @@ def cli(self, vrf_type='all', vrf='all', instance='all', neighbor='', address_fa
                     sub_dict['link_state'] = m.groupdict()['link_state']
                 sub_dict['local_as_as_no'] = int(m.groupdict()['local_as_as_no'])
 
-                # Default the values - overwritten if configured
-                sub_dict['local_as_no_prepend'] = False
-                sub_dict['local_as_replace_as'] = False
-                sub_dict['local_as_dual_as'] = False
+                if m.groupdict()['local_as_no_prepend']:
```
tahigash, line 2674: thanks for fixing/updating this. Can I ask you to add UT with output from device?

ancker010: Am I mistaken that there's currently no UT or folder based test at all for this command?

tahigash: there is as old style UT.
https://raw.githubusercontent.com/CiscoTestAutomation/genieparser/master/src/genie/libs/parser/iosxr/tests/test_show_bgp.py

But if you can, please add UT for this as folder based test.

ancker010: I've updated the old-style test, but I really don't have the capacity to make a whole new folder based test for this command.

ancker010: Update. I think I've got this done. The folder-based test works on my local system.

Screen Shot 2021-08-03 at 11 03 04 AM

tahigash: thanks. yes, only what you developed is fine.

tahigash (approved):
thanks for addressing comments and contribution.

lsheikal (approved):

lsheikal:
Thank you for your contribution!!!

PR #506, opened 2021-08-13
IOSXE: ShowIpRouteSummary - Allow ISIS to use no/null area-tag

oianson:
## Description
With ISIS, you can configure ISIS without an area-tag or with an area tag (Test in example below)

```
#show ip route summary
IP routing table name is default (0x0)
IP routing table maximum-paths is 32
Route Source    Networks    Subnets     Replicates  Overhead    Memory (bytes)
isis Test       0           0           0           0           0
  Level 1: 0 Level 2: 0 Inter-area: 0
isis            0           0           0           0           0
  Level 1: 0 Level 2: 0 Inter-area: 0
```
Currently if there is no area-tag set 'instance' doesn't exist and therefore gives an error when it comes to update it here: `vrf_rs_dict['isis'][instance].update(group)`

Not sure if this solution is a good one though to allow for null area-tags. Please provide feedback 

## Checklist:


LukasMcClelland (approved):
Thank you for the contribution! Looks great 👍

lsheikal (approved):

PR #513, opened 2021-08-25
add 'show ipv6 interface' to 'ShowIpv6Interface' class

oianson:
## Description

* IOSXR
    * Modified ShowIpv6Interface:
        * Added 'show ipv6 interface' as a CLI command for this parser

## Motivation and Context
Got a no parser found error without this

## Checklist:


src/genie/libs/parser/iosxr/show_interface.py
```diff
@@ -1576,7 +1576,8 @@ class ShowIpv6VrfAllInterface(ShowIpv6VrfAllInterfaceSchema):
     """Parser for show ipv6 vrf all interface"""
 
     cli_command = ['show ipv6 vrf {vrf} interface {interface}',
-                   'show ipv6 vrf {vrf} interface', 'show ipv6 vrf all interface']
+                   'show ipv6 vrf {vrf} interface', 'show ipv6 vrf all interface',
+                   'show ipv6 interface']
```
LukasMcClelland, line 1580: Please add logic with the `cli` function to handle this new command. See screenshot below of the `if` block that determines which command to send to the device.

image


You _might_ be able to simply add in an `elif not interface and not vrf:`, but if that doesn't work, then you'll have to make a small subclass for the 'show ipv6 interface' command

LukasMcClelland (changes requested):
Thank you for contributing!

Please add [folder-based test files](https://pubhub.devnetcloud.com/media/pyats-development-guide/docs/writeparser/writeparser.html#folder-based-testing-asa-ios-and-iosxe) for your command too

src/genie/libs/parser/iosxr/show_interface.py
oianson: Isn't this already covered by the else on line 1592? It will be used when there is no VRF or interface specified. 

```
            else:
                cmd = self.cli_command[2]
```

oianson: Hmm. I'm not sure how to work with the if block for `show ipv6 vrf all interface` and `show ipv6 interface`  as neither will have a vrf or interface parameter specified. 

So currently `show ipv6 interface` would be hitting the else statement on line 1592/1593, but it would be running `show ipv6 vrf all interface`

oianson: There was already a class for ShowIpv6Interface .. I added the cli command there instead.. think thats all I need to do ..?

PR #516, opened 2021-08-27
IOSXR/NXOS: fix "show static routing" parsers to correctly handle IPv6

vincentbernat:
## Description

The regex to capture IPv6 were incorrect. I tested this on ASR9k,
NCS5k and NX7k. The tests were passing due to IPv6 addresses only
having digits. I have updated them to add non-digits. I have checked
there is no other occurences in the code base, but just to be sure, I
will push a followup commits to update other tests to add a non-digit
in IPv6 addresses.

## Checklist:

### DIFF
--- a/changelog/undistributed/changelog_iosxr_showStaticTopologyDetail_20210827.rst
+++ b/changelog/undistributed/changelog_iosxr_showStaticTopologyDetail_20210827.rst
@@ -1,0 +1,9 @@
+--------------------------------------------------------------------------------
+                            Fix
+--------------------------------------------------------------------------------
+* IOSXR
+    * Modified ShowStaticTopologyDetail:
+        * Correctly match IPv6 addresses
+* NXOS
+    * Modified ShowIpv6StaticRoute
+        * Correctly match IPv6 addresses
--- a/src/genie/libs/parser/ios/tests/ShowBgpAllDetail/cli/equal/golden_output1_expected.py
+++ b/src/genie/libs/parser/ios/tests/ShowBgpAllDetail/cli/equal/golden_output1_expected.py
@@ -155,7 +155,7 @@ expected_output = {
                         },
                         "ipv6 unicast": {
                             "prefixes": {
-                                "2001:1:1:1::1/128": {
+                                "2001:1:1:a::1/128": {
                                     "available_path": "1",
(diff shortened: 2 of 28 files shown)


GerriorL:
Looks like you'll need to merge the latest code into your branch before it can be tested.

```
git remote add upstream https://github.com/CiscoTestAutomation/genieparser.git
git fetch upstream
git pull upstream
```

GerriorL (approved):

domachad (changes requested):
Hi Vincent, thank you for your contribution! 
There is just one important thing that needs to be done before we can approve these changes. We are moving away from unittest based testing and using folder-based tests instead.

Can you add folder-based tests that cover your changes over these commands? 
If you pull the latest changes in genieparser, you will notice that some (if not all) of these test files no longer exist.

vincentbernat:
I am not adding new tests. I am only modifying existing tests that failed to detect the problem due to IPv6 only using digits.

domachad:
@vincentbernat 

Unfortunately, I believe I'll have to wait until after the monthly release is over (which will happen this week) to properly review your PR as we're making sweeping changes to parser tests (NXOS tests included). After release you'll be able to see what changes have been made and my comment about the files moving will make more sense.

As for changing existing tests, the common golden rule we follow is to maintain 100% backwards compatibility (when possible) and for parsers this equates to leaving all existing unit tests alone. However, in this case it does seem like overkill to ask for a new 1200 line test file for a change as small as this. I'll check with my team and see what we can do

vincentbernat:
The change itself is super simple: fix the regex to parse IPv6 address. The current regex fails to parse an IPv6 address like 2001:db8::14. This bug is not caught by current tests because all tests are using IPv6 addresses only containing digits. I am modifying a lot of existing tests to ensure such a faulty regex is not present elsewhere.

I have already rebased this PR once because of the previous monthly release moving tests from unittests to folder-based tests, I won't rebase it a second time. The monthly single commit release is a bit rude to contributors like me using the regular pull request workflow.

domachad:
@vincentbernat the conflicts were handled internally and your code was merged. 
I will close this pull request.
Your changes will be available as part of the next release.
Thank you again for your contribution

PR #523, opened 2021-09-02
added make json to github actions

tahigash:
## Description
added `make json` to github actions

## Motivation and Context
no more failure because of not-updated json file

## Impact (If any)
n/a

## Screenshots:
see github action result

## Checklist:

### DIFF
--- a/.github/workflows/run_tests.yml
+++ b/.github/workflows/run_tests.yml
@@ -32,6 +32,8 @@ jobs:
     - name: Folder based testing
       run: |
         make develop
+        make json
+        git status
         cd tests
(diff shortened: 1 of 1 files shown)


d-the-iii (approved):
LGTM, are you sure this doesn't need a changelog though? This seems like a change worth recording somewhere.

LukasMcClelland (approved):

tahigash:
> LGTM, are you sure this doesn't need a changelog though? This seems like a change worth recording somewhere.

we usually don't add pipeline update to changelog. so should be fine. still we can track from git history if we want.

Issue #515, opened 2021-08-26
Junos - "show interfaces" - UnboundLocalError "address_family_dict"


Kani999:
Hello,

The parser `show interfaces` for my device with OS `junos`  fails with the following error

```
2021-08-26 13:11:56,869: %UNICON-INFO: connection to device
--- JUNOS 15.1R4.6 built 2016-06-23 21:29:57 UTC
2021-08-26 13:12:01,698: %UNICON-INFO: +++ device with alias 'cli': executing command 'show interfaces' +++
show interfaces
Physical interface: ge-0/0/0, Enabled, Physical link is Up
  Interface index: 130, SNMP ifIndex: 502
.
.   Interfaces output
.

Physical interface: vme, Enabled, Physical link is Down
.
.   Interfaces output
.
{master:0}
user@device >

Traceback (most recent call last):
  File "test.py", line 11, in <module>
    p1 = dev.parse('show interfaces')
  File "src/genie/conf/base/device.py", line 523, in genie.conf.base.device.Device.parse
  File "src/genie/conf/base/device.py", line 550, in genie.conf.base.device.Device._get_parser_output
  File "src/genie/metaparser/_metaparser.py", line 272, in genie.metaparser._metaparser.MetaParser.parse
  File "/home/jankrupa/.local/lib/python3.8/site-packages/genie/libs/parser/junos/show_interface.py", line 1962, in cli
    address_family_flags_dict = address_family_dict.setdefault('address-family-flags', {})
UnboundLocalError: local variable 'address_family_dict' referenced before assignment
```

tahigash:
@Kani999, thanks for reporting the issue. genie parser is open-sourced. so we are encouraging users to contribute fixes/enhancement like above. can you consider contributing above fix?

references:
https://pubhub.devnetcloud.com/media/pyats-development-guide/docs/writeparser/writeparser.html
https://pubhub.devnetcloud.com/media/pyats-development-guide/docs/contribute/contribute.html#contribute

tahigash:
thanks for opening the PR and contribution! We now can track on the PR. So, I'm closing this issue. but feel free to open when needed.

PR #528, opened 2021-09-03
Initiate empty dict before referencing variable

Kani999:
Fixes: #515

## Description
Variable `address_family_dict` was referenced before assignment;

## Motivation and Context
Parser causes error:

```
Traceback (most recent call last):
  File "junos_show_interfaces_sample.py", line 11, in <module>
    res = dev.parse("show interfaces", output=output)
  File "src/genie/conf/base/device.py", line 524, in genie.conf.base.device.Device.parse
  File "src/genie/conf/base/device.py", line 557, in genie.conf.base.device.Device._get_parser_output
  File "src/genie/metaparser/_metaparser.py", line 272, in genie.metaparser._metaparser.MetaParser.parse
  File "/home/jankrupa/genie_fix/genieparser/src/genie/libs/parser/junos/show_interface.py", line 1963, in cli
    address_family_flags_dict = address_family_dict.setdefault('address-family-flags', {})
```

After assigning an empty dict to `address_family_dict` the output is return in proper JSON format.

## Checklist:


src/genie/libs/parser/junos/show_interface.py
```diff
@@ -1480,6 +1480,7 @@ def cli(self, interface=None, output=None):
         lag_int_type = ''
         lacp_flag = ''
         if_dist_dict = {}
+        address_family_dict = {}
```
LukasMcClelland, line 1483: Good catch

I would recommend also changing line 1938 to use `.update` like on line 1946 so that we don't overwrite the dictionary; only add to it/update it
image

LukasMcClelland (changes requested):
Thank you for contributing the fix!

Please add a changelog as well. This repo's README has instructions on how to do that

src/genie/libs/parser/junos/show_interface.py
Kani999: @LukasMcClelland I hope it's fine now

LukasMcClelland:
Hi @Kani999 

Looks like these changes are causing existing parser tests to fail 
image

You can run these tests yourself using the `folder_parsing_job.py` script in the `tests` folder. Docs are [here](https://pubhub.devnetcloud.com/media/pyats-development-guide/docs/writeparser/writeparser.html#folder-based-testing-asa-ios-and-iosxe) for that

Please adjust the code so that these tests pass. If you need to remove my suggested change, then so be it 👍 

Thanks

Kani999:
> Hi @Kani999
> 
> Looks like these changes are causing existing parser tests to fail
> 
> Please adjust the code so that these tests pass. If you need to remove my suggested change, then so be it 👍
> 
> Thanks

I've looked at the tests but I cannot find what causes the problem.
I've changed the `junos/show_interface.py:1934` line back to its original state and the tests pass fine.

LukasMcClelland:
Looks like the tests are passing 👍 

Even though this is a very small change, it would still be great to have a test for it. It will help to ensure that the parser will continue to parse your devices output when any future changes are made to the file. 

Please add a test and feel free to use the output from the issue you created
```
Physical interface: ge-0/0/0, Enabled, Physical link is Up
  Interface index: 130, SNMP ifIndex: 502
.
.   Interfaces output
.

Physical interface: vme, Enabled, Physical link is Down
.
.   Interfaces output
.
{master:0}
```

Link to documentation for writing parser tests -> https://pubhub.devnetcloud.com/media/pyats-development-guide/docs/writeparser/writeparser.html#folder-based-testing-asa-ios-and-iosxe

LukasMcClelland (changes requested):
As mentioned just above, please include a test

ThomasJRyan (approved):

PR #529, opened 2021-09-03
IOSXR: fix "show route ipv4" parser for hops without interface

vincentbernat:
"show route ipv4" may not provide an interface for each hop, like in
this example:

```
B    10.0.0.0/8 [20/30] via 172.19.165.189, 3w0d
                [20/0] via 172.30.254.254, 3w0d
```

Fix the regex to handle this case correctly. Previously, `3w0d` was
considered to be the interface. The regex was stolen from the IPv6
case which was correct (and tested).

## Checklist:


src/genie/libs/parser/iosxr/show_routing.py
```diff
@@ -586,9 +586,10 @@ def cli(self, vrf=None, route=None, protocol=None, output=None):
 
         # [90/15360] via 10.23.90.3, 1w0d, GigabitEthernet0/0/0/1.90
         # [110/2] via 10.1.2.1, 01:50:49, GigabitEthernet0/0/0/3
-        p3 = re.compile(r'^\[(?P<route_preference>\d+)\/(?P<metric>\d+)\] +via +'
-                        r'(?P<next_hop>\S+),( +(?P<date>[\w:]+))?,? +'
-                        r'(?P<interface>[\w\/\.\-]+)$')
+        # [110/2] via 10.1.3.1, 3w3d
+        p3 = re.compile(r'^\[(?P<route_preference>\d+)\/(?P<metric>\d+)\] +'
```
tahigash, line 590: thanks for contribution for this fix! can I ask you to add testing for this change by adding device output with parsed output?

reference:
https://pubhub.devnetcloud.com/media/pyats-development-guide/docs/writeparser/writeparser.html#testing-your-parser

tahigash (changes requested):
thanks for the fix. left one comment.

d-the-iii (commented):
Fix looks good to me, but I agree with Takashi. Do you have an example of this line without the interface so we can manually test the regex, and to make unit tests with?

src/genie/libs/parser/iosxr/show_routing.py
vincentbernat: I have added a test for this.

tahigash (approved):
thanks for addressing comments and contribution!

d-the-iii (approved):

PR #530, opened 2021-09-03
IOSXR: fix AS parsing for dynamic neighbors

vincentbernat:
Dynamic neighbors get an additional "Dynamic" property on the line
where remote and local AS are specified:

```
Remote AS 65498, local AS 65499, external link, Dynamic
```

Currently, the regex to parse this line expects a precise set of
properties and reject any unkown property. I propose to extend the
regex to accept and discard any unknown property. It seems better than
to return neighbors without local/remote AS, as this is quite an
essential information.

## Checklist:


GerriorL:
Perhaps it would be better to capture it and add it to the schema as an Optional() key. That way it will get captured and recorded if it's there and if it's not there the schema would continue as normal. Either way you're correct that the regex needs to be able to handle additional properties. Thank you for catching this!

danielgraziano (approved):
Please address the comment @GerriorL left then its good

PR #536, opened 2021-09-07
Updated iosxe show license summary regex pattern

dannywade:
## Description
Updated the regex pattern for the license capture group to accommodate various outputs.

## Motivation and Context
I discovered that the license name may vary, so I wanted to make the regex pattern more broad.

## Impact (If any)
N/A

## Screenshots:
licensesumm-fix-make

licensesumm-fix-test_output

## Checklist:

- [X] I have updated the changelog.
- [X] I have updated the documentation (If applicable).
- [X] I have added tests to cover my changes (If applicable).
- [X] All new and existing tests passed.

### DIFF
--- a/src/genie/libs/parser/iosxe/show_license.py
+++ b/src/genie/libs/parser/iosxe/show_license.py
@@ -217,7 +217,8 @@ class ShowLicenseSummary(ShowLicenseSummarySchema):
 
         # network-advantage       (C9300-48 Network Advan...)       1 IN USE
         # dna-advantage           (C9300-48 DNA Advantage)          1 IN USE
-        p0 = re.compile(r"^(?P<license>[\w-]+)\s+\((?P<entitlement>.+)\)\s+(?P<count>\d+)\s+(?P<status>.+)$")
+        
+        p0 = re.compile(r"^(?P<license>.+\S+)\s+\((?P<entitlement>.+)\)\s+(?P<count>\d+)\s+(?P<status>.+)$")
 
         for line in out.splitlines():
             line=line.strip()
--- a/src/genie/libs/parser/iosxe/show_license.py
+++ b/src/genie/libs/parser/iosxe/show_license.py
@@ -218,12 +218,7 @@ class ShowLicenseSummary(ShowLicenseSummarySchema):
         # network-advantage       (C9300-48 Network Advan...)       1 IN USE
         # dna-advantage           (C9300-48 DNA Advantage)          1 IN USE
 
-        # Example 2
         # C9300 48P DNA Advantage  (C9300-48 DNA Advantage)          2 AUTHORIZED
-        # C9300 48P Network Adv... (C9300-48 Network Advan...)       2 AUTHORIZED
-        # C9300 24P Network Adv... (C9300-24 Network Advan...)       1 AUTHORIZED
-        # C9300 24P DNA Advantage  (C9300-24 DNA Advantage)          1 AUTHORIZED
-        
         p0 = re.compile(r"^(?P<license>.+?)\s+\((?P<entitlement>.+)\)\s+(?P<count>\d+)\s+(?P<status>.+)$")
 
         for line in out.splitlines():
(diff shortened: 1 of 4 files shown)


src/genie/libs/parser/iosxe/show_license.py
```diff
@@ -217,7 +217,8 @@ def cli(self, output=None):
 
         # network-advantage       (C9300-48 Network Advan...)       1 IN USE
         # dna-advantage           (C9300-48 DNA Advantage)          1 IN USE
-        p0 = re.compile(r"^(?P<license>[\w-]+)\s+\((?P<entitlement>.+)\)\s+(?P<count>\d+)\s+(?P<status>.+)$")
+        
+        p0 = re.compile(r"^(?P<license>.+\S+)\s+\((?P<entitlement>.+)\)\s+(?P<count>\d+)\s+(?P<status>.+)$")
```
d-the-iii, line 221: Can you provide an example of a line this will now parse that the previous one didn't? Additionally, can you update the unit tests to use that example?

Also, the `.+\S+` seems odd to me. `.+` will already capture one or more of any character. Isn't it redundant to have `\S+` after? Anything that would be captured by `\S+` will also be captured by `.+`

d-the-iii (changes requested):

src/genie/libs/parser/iosxe/show_license.py
dannywade: I'll add an example of different output and update the unit tests with the new example.

In the new example, you'll see that there's some whitespace in the license names. The idea behind the regex was to capture all the characters in the license name (whitespace included), but not capture the whitespace after it (between the license name and entitlement tag). I'm open to a better suggestion for regex, as it probably can be done in a more efficient manner.

d-the-iii: I understand. My suggested solution would be to change the license capture group to this `(?P<license>.+?)`
The full line would be
```
^(?P<license>.+?)\s+\((?P<entitlement>.+)\)\s+(?P<count>\d+)\s+(?P<status>.+)$
```
The `?` after the plus will make its behavior lazy instead of greedy, so it will capture up to the whitespace, and then the \s+ regex outside the capture group will take over.

dannywade: Perfect! That's exactly what I was looking for. Thank you for the suggestion. Changes have been made to the regex pattern.

src/genie/libs/parser/iosxe/show_license.py
```diff
@@ -217,7 +217,14 @@ def cli(self, output=None):
...
         # network-advantage       (C9300-48 Network Advan...)       1 IN USE
         # dna-advantage           (C9300-48 DNA Advantage)          1 IN USE
-        p0 = re.compile(r"^(?P<license>[\w-]+)\s+\((?P<entitlement>.+)\)\s+(?P<count>\d+)\s+(?P<status>.+)$")
+
+        # Example 2
+        # C9300 48P DNA Advantage  (C9300-48 DNA Advantage)          2 AUTHORIZED
+        # C9300 48P Network Adv... (C9300-48 Network Advan...)       2 AUTHORIZED
+        # C9300 24P Network Adv... (C9300-24 Network Advan...)       1 AUTHORIZED
+        # C9300 24P DNA Advantage  (C9300-24 DNA Advantage)          1 AUTHORIZED
+
```
d-the-iii, line 226: Small couple of changes to this section. 
Drop the Example 2 line. 
Remove the blank line between the examples and the pattern.
Drop the other examples and leave only the top one. They're all very similar so only one example is needed.

Should look like this
```suggestion
        # C9300 48P DNA Advantage  (C9300-48 DNA Advantage)          2 AUTHORIZED
```

d-the-iii (approved):
I'm gonna go ahead and approve, just change the examples bit to match how we normally do our examples. I have placed a comment below explaining.

domachad (approved):

PR #544, opened 2021-09-17
Updated ShowIpRouteSummary

GerriorL:
## Description
Added in a second regex and a switch to capture devices with a different show output. Schema Optionals added.

## Motivation and Context
https://github.com/CiscoTestAutomation/pyats/issues/129

## Checklist:

### DIFF
--- a/src/genie/libs/parser/iosxe/show_routing.py
+++ b/src/genie/libs/parser/iosxe/show_routing.py
@@ -1958,11 +1958,11 @@ class ShowIpRouteSummary(ShowIpRouteSummarySchema):
                         protocol_dict = vrf_dict.setdefault('total_route_source', {})
                     else:
                         protocol_dict = vrf_rs_dict.setdefault(protocol, {})
-                    if instance is not None:
+                    if instance:
                         inst_dict = protocol_dict.setdefault(instance, {})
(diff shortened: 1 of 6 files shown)


src/genie/libs/parser/iosxe/show_routing.py
```diff
@@ -1916,23 +1939,51 @@ def cli(self, vrf='', output=None):
...
-                continue
+                if replicates_flag:
+                    group = m.groupdict()
+                    protocol = group.pop('protocol')
+                    instance = group.pop('instance')
+                    if protocol == 'Total':
+                        protocol_dict = vrf_dict.setdefault('total_route_source', {})
+                    else:
+                        protocol_dict = vrf_rs_dict.setdefault(protocol, {})
+                    if instance is not None:
```
domachad, line 1961: if instance does not do the same thing in this case?
```suggestion
                    if instance:
```

domachad (approved):
Approved with a suggestion

src/genie/libs/parser/iosxe/show_routing.py
GerriorL: That's correct. I'll update it to be consistent with my additions

tahigash (approved):

Issue #569, opened 2021-11-03
Data loss when parsing ASA `show route` output


thismatters:
Hello,

I'm parsing some `show route` output out in the wild, and I stumbled across what seems like a limitation of the schema.

This is an excerpt from `show route` on "device-a":

```
S*       0.0.0.0 0.0.0.0 [1/0] via 192.2.0.1, Outside-Port-Channel
...
S    0.0.0.0 0.0.0.0 [255/0] via 10.98.126.1, Inside-Port-Channel tunneled 
```

The parsed output from this config omits the top row because the CIDR coincides with the bottom row:

```
...
"routes": {
    "0.0.0.0/0": {
        "candidate_default": false,
        "active": true,
        "route": "0.0.0.0/0",
        "source_protocol_codes": "S",
        "source_protocol": "static",
        "metric": 0,
        "route_preference": 255,
        "next_hop": {
            "next_hop_list": {
                1: {
                    "index": 1,
                    "next_hop": "10.98.126.1",
                    "outgoing_interface_name": "Inside-Port-Channel"
                }
            }
        }
    },
...
}
...
```

I propose that the [schema for ASA's show route](https://github.com/CiscoTestAutomation/genieparser/blob/master/src/genie/libs/parser/asa/show_route.py#L18) (and possibly [others](https://github.com/CiscoTestAutomation/genieparser/blob/master/src/genie/libs/parser/iosxe/show_routing.py#L113)) be updated to support plural routes:

```
...
"routes": {
    "0.0.0.0/0": {
        1: {
            "candidate_default": true,
            "active": true,
            "route": "0.0.0.0/0",
            "source_protocol_codes": "S",
            "source_protocol": "static",
            "metric": 0,
            "route_preference": 1,
            "next_hop": {
                "next_hop_list": {
                    "1": {
                        "index": 1,
                        "next_hop": "192.2.0.1",
                        "outgoing_interface_name": "Outside-Port-Channel"
                    }
                }
            }
            }
        },
        2: {
            "candidate_default": false,
            "active": true,
            "route": "0.0.0.0/0",
            "source_protocol_codes": "S",
            "source_protocol": "static",
            "metric": 0,
            "route_preference": 255,
            "next_hop": {
                "next_hop_list": {
                    1: {
                        "index": 1,
                        "next_hop": "10.98.126.1",
                        "outgoing_interface_name": "Inside-Port-Channel"
                    }
                }
            }
        }
    }
}
...
```

[...]

PR #572, opened 2021-11-03
schema change

thismatters:
## Description
Changed the schema and parser for the ASA `show route` command to permit multiple routes with the same subnet.

## Motivation and Context

```
S*       0.0.0.0 0.0.0.0 [1/0] via 104.225.199.193, Outside-Port-Channel
```

The output of the parser omits the first row while including the second. This is problematic because the first row represents the route that traffic would follow by default under normal conditions in the actual network. Ergo the parser output doesn't accurately reflect the config.

## Impact (If any)
There is considerable impact to the users of the library due to the schema change. I'm not sure what the policies are around schema changes and how they are scheduled in the roadmap.

## Screenshots:
Haven't done compile/demo yet

## Checklist:

- [ ] I have updated the documentation (If applicable). -- N/A

### DIFF
--- a/changelog/2021/november.rst
+++ b/changelog/2021/november.rst
@@ -1,0 +1,9 @@
+--------------------------------------------------------------------------------
+                                      Fix
+--------------------------------------------------------------------------------
+
+* asa
+    * Modified ShowRouteSchema
+        * Added index to routes and supporting multiple routes to same subnet
+    * Modified ShowRoute
+        * Added index to routes and supporting multiple routes to same subnet
--- a/src/genie/libs/parser/asa/show_route.py
+++ b/src/genie/libs/parser/asa/show_route.py
@@ -26,26 +26,29 @@ class ShowRouteSchema(MetaParser):
                     'ipv4': {
                         Optional('routes'): {
                             Any(): {
-                                'candidate_default': bool,
-                                Optional('subnet'): str,
-                                'route': str,
-                                Optional('active'): bool,
-                                Optional('date'): str,
-                                Optional('route_preference'): int,
-                                Optional('metric'): int,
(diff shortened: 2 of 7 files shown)


LukasMcClelland:
>There is considerable impact to the users of the library due to the schema change. I'm not sure what the policies are around schema changes and how they are scheduled in the roadmap.

I'm pleased that you appreciate that. The good and bad answer is that all changes need to be backwards compatible. We can't break someone's tests if this parser has been included into their scripts over the past 2 year since it was released. This also means that all existing tests have to be unchanged and still pass with the changes you'll make. 

This enhancement isn't impossible, but it will require some of the schema to be repeated and many keys will end up being Optional. [This schema for ShowVrrp](https://github.com/CiscoTestAutomation/genieparser/blob/158f56a25dde2bbb71bf01f7ea0d0f34c93e3590/src/genie/libs/parser/iosxe/show_vrrp.py#L44) is an example

thismatters:
> The good and bad answer is that all changes need to be backwards compatible. 

Commendable.

> This enhancement isn't impossible, but it will require some of the schema to be repeated and many keys will end up being Optional. [This schema for ShowVrrp](https://github.com/CiscoTestAutomation/genieparser/blob/158f56a25dde2bbb71bf01f7ea0d0f34c93e3590/src/genie/libs/parser/iosxe/show_vrrp.py#L44) is an example

I'll check that out.

thismatters:
I put an idea in the issue (#569) for incorporating this change without changing the schema whatsoever; although it does (in the edge case concerned) change the parsed output. Let me know if it satisfies the backward compatibility constraints.

LukasMcClelland (commented):
Left a comment on the issue you linked

PR #577, opened 2021-11-18
Added parser for 'show access-list' and 'show access-list {id} to ASA

MikeJames15:
## Checklist:


MikeJames15:
This is my first time contributing. Please let me know if I messed something up. I tried to follow all the instructions.
I had trouble running the tests. Couldn't seem to be able to install cisco-distutils.

danielgraziano:
> This is my first time contributing. Please let me know if I messed something up. I tried to follow all the instructions. I had trouble running the tests. Couldn't seem to be able to install cisco-distutils.

No worries, the github checks run the tests and they seem to have passed

changelog/undistributed/changelog_show_access_list_20210614202815.rst
```diff
@@ -0,0 +1,8 @@
+--------------------------------------------------------------------------------
+                                New
+--------------------------------------------------------------------------------
+* ASA
+    * Added ShowAccessList:
+      * show access-list
```
danielgraziano, line 6: Please make sure the indentation is 4 spaces per indent. During the release process we use an automated script to generate the final changelog

src/genie/libs/parser/asa/show_access_list.py
```diff
@@ -0,0 +1,187 @@
...
+class ShowAccessListSchema(MetaParser):
+    """Schema for
+        * show access-list
+    """
+
+    schema = {
+    	'access-list': {
+            Any(): {
+                Optional('elements'): int,
+#               'name_hash': str,
```
danielgraziano, line 27: please remove if not needed

src/genie/libs/parser/asa/show_access_list.py
```diff
@@ -0,0 +1,187 @@
...
+    """
+
+    cli_command = 'show access-list'
+
+    def cli(self, output=None):
+        if output is None:
+            # excute command to get output
+            out = self.device.execute(self.cli_command)
+        else:
+            out = output
```
danielgraziano, line 76:
``` python
if output is None:
    output = self.device.execute(self.cli_command)

(snip)

for line in output.splitlines():

File: src/genie/libs/parser/asa/show_access_list.py [hunk 4]
```diff
@@ -0,0 +1,187 @@
...
+
+        ret_dict = {}
+        ret_dict['access-list'] = {}
```
danielgraziano, line 79: Please remove this. We only want to add to the dictionary within the for-loop. This way if nothing is parsed an empty dictionary will get returned (we expect this behaviour)

src/genie/libs/parser/asa/show_access_list.py
```diff
@@ -0,0 +1,187 @@
...
+            if line == '':
+                continue
+
+            m = p1.match(line)
+            if m:
+                groups = m.groupdict()
+                acl_name = groups['name']
+                ret_dict['access-list'][acl_name] = {}
+                ret_dict['access-list'][acl_name]['elements'] = int(groups['elements'])
+#                ret_dict['access-list'][acl_name]['name_hash'] = groups['namehash']
```
danielgraziano, line 116: please remove if not used

danielgraziano: But im curious. How come you are not parsing this?

src/genie/libs/parser/asa/show_access_list.py
```diff
@@ -0,0 +1,187 @@
...
+            line = line.strip()
+
+            if line == '':
+                continue
+
+            m = p1.match(line)
+            if m:
+                groups = m.groupdict()
+                acl_name = groups['name']
+                ret_dict['access-list'][acl_name] = {}
```
danielgraziano, line 114:
``` python
acl_dict = ret_dict.setdefault('access-list', {}).setdefault(acl_name, {})

File: src/genie/libs/parser/asa/show_access_list.py [hunk 7]
```diff
@@ -0,0 +1,187 @@
...
+                ret_dict['access-list'][acl_name]['elements'] = int(groups['elements'])
```
danielgraziano, line 115:
``` python
acl_dict.update({'elements': int(groups['elements']})

File: src/genie/libs/parser/asa/show_access_list.py [hunk 8]
```diff
@@ -0,0 +1,187 @@
...
+            
+            m = p2.match(line)
+            if m:
+                groups = m.groupdict()
+                acl_name = groups['name']
+                line_num = groups['entry']
+                if not acl_name in ret_dict['access-list'].keys() :
+                    ret_dict['access-list'][acl_name] = {}
+                if not 'entry' in ret_dict['access-list'][acl_name].keys():
+                    ret_dict['access-list'][acl_name]['entry'] = {}
```
danielgraziano, line 127: The magic of `<dict>.setdefault()` is it wont overwrite existing keys. It will only insert if the key doesnt exist.
``` python
acl_entry_dict = ret_dict.setdefault('access-list', {}).setdefault(acl_name, {}).setdefault('entry', {})

File: src/genie/libs/parser/asa/show_access_list.py [hunk 9]
```diff
@@ -0,0 +1,187 @@
...
+                continue
+
+            m = p3.match(line)
+            if m:
+                groups = m.groupdict()
+                line_num = groups['entry']
+                if not acl_name in ret_dict['access-list'].keys() :
+                    ret_dict['access-list'][acl_name] = {}
+                if not 'entry' in ret_dict['access-list'][acl_name].keys():
+                    ret_dict['access-list'][acl_name]['entry'] = {}
```
danielgraziano, line 148: same with setdefault

src/genie/libs/parser/asa/show_access_list.py
```diff
@@ -0,0 +1,187 @@
...
+                ret_dict['cached_acl_log_flows'] = flows
+                continue
+
+
+            m = p5.match(line)
+            if m:
+                groups = m.groupdict()
+                ret_dict['alert_interval'] = int(groups['alert_interval'])
+#            else:
+#                print('Not matching line: {}'.format(line))
```
danielgraziano, line 170: please remove

src/genie/libs/parser/asa/show_access_list.py
```diff
@@ -0,0 +1,187 @@
...
+
+
+        if not(bool(ret_dict['access-list'])):
+            ret_dict = {}
```
danielgraziano, line 174: please remove, not needed if the other comments are updated

src/genie/libs/parser/asa/show_access_list.py
```diff
@@ -0,0 +1,187 @@
...
+        for name in names:
+            if name in groups and not groups[name] is None:
+                line_dict[name] = groups[name]
+
+    def add_host(self, line_dict, groups, prefix):
+        names = ['host', 'any', 'object_group']
+        for name in names:
+            prefixed_name = prefix + name
+            if prefixed_name in groups and not groups[prefixed_name] is None:
+                line_dict[name] = groups[prefixed_name]
```
danielgraziano, line 187: Please do this logic within each regex match. There shouldnt be any more methods defined so that all parsers follow the same format

danielgraziano (changes requested):

src/genie/libs/parser/asa/show_access_list.py
MikeJames15: Added the value to the output

src/genie/libs/parser/asa/show_access_list.py
```diff
@@ -0,0 +1,309 @@
+''' show_access_list.py
+
+Parser for the following show commands:
+    * show access-list
+'''
+
+# Python
+import re
+import logging
+from collections import OrderedDict
```
danielgraziano, line 10: No need to use an OrderedDict, please use a regular dictionary

src/genie/libs/parser/asa/show_access_list.py
```diff
@@ -0,0 +1,309 @@
+''' show_access_list.py
+
+Parser for the following show commands:
+    * show access-list
+'''
+
+# Python
+import re
+import logging
```
danielgraziano, line 9: Not used, please remove

src/genie/libs/parser/asa/show_access_list.py
```diff
@@ -0,0 +1,309 @@
...
+        previous_line_num = -1
+        previous_acl_name = ''
+        line_sub = -1
+        for line in output.splitlines():
+            line = line.strip()
+
+            if line == '':
+                continue
+
+            m = p1.match(line)
```
danielgraziano, line 222: please add a comment above each regex usage that shows what lines the regex can match (same ones from above). Please do below as well

danielgraziano (changes requested):
Should be good after these changes

src/genie/libs/parser/asa/show_access_list.py
MikeJames15: In access lists it makes a difference in which order the entries appear, for example when looping over them it is not the same if we use a regular dict or an OrderedDict. An OrderedDict will provide them in the right order and a regular dict will not guarantee it.

src/genie/libs/parser/asa/show_access_list.py
MikeJames15: I have removed the logging import

src/genie/libs/parser/asa/show_access_list.py
MikeJames15: Copied the comments from the pattern definitions to the match lines such as "m = p1.match(line)"

MikeJames15:
My tests have started to fail without making any changes ( before I did my last commit  and without any changes from when they were working). What should I do?

MikeJames15:
I have some changes to show_bgp.py for iosxe. 

The changes include:

- adding up_time, down_time, last_read and last_write to the neighbor level. these values were appearing under address_family in the schema for some reason even though they are not dependent on the address_family and they were not getting set.

- fixed bug where split lines were not processed. When a line ends after the prefix, it is saved and joined with the following one.

- added support for AS with a dot in them (65324.3) in the AS path. Only managed to do this in one of the regexes (p6) the other one was more complicated to implement. May do it in the future.

- used setdefault() instead of an if followed by creation of dicts, as recommended for show_access_list.py.

- removed empty address_family entry created when only data for a RD is available.

Have not pushed this change yet. Should I add it to this pull request, wait for this to be resolved, add it to another branch in my fork or do a separate fork for it?

Please let me know how to continue.

PR #582, opened 2021-11-26
569 asa show route take2

thismatters:
## Description
Changed the parser for the ASA show route command to permit multiple routes with the same subnet.

## Motivation and Context

```
S*       0.0.0.0 0.0.0.0 [1/0] via 192.2.0.1, Outside-Port-Channel
```

The output of the parser omits the first row while including the second. This is problematic because the first row represents the route that traffic would follow by default under normal conditions in the actual network. Ergo the parser output doesn't accurately reflect the config.

## Impact (If any)
Entries with a duplicated route will not be overwritten. Users who have been experiencing quiet data loss will have divergent output from prior versions.

## Checklist:
- [ ] I have updated the documentation (If applicable). N/A


src/genie/libs/parser/asa/tests/ShowRoute/cli/equal/golden_output_4_expected.py
```diff
@@ -0,0 +1,84 @@
...
+                            "next_hop": {
+                                "next_hop_list": {
+                                    1: {
+                                        "index": 1,
+                                        "next_hop": "10.98.126.1"
+                                    }
+                                }
+                            }
+                        },
+                        "0.0.0.0/0-1": {
```
lsheikal, line 61: This seems weird. I would suggest to use index here.

src/genie/libs/parser/asa/tests/ShowRoute/cli/equal/golden_output_4_expected.py
```diff
@@ -0,0 +1,84 @@
+expected_output = {
+    "vrf": {
+        "default": {
+            "address_family": {
+                "ipv4": {
+                    "routes": {
+                        "0.0.0.0/0": {
```
lsheikal, line 7:
```suggestion
                    "routes": {
                         0: {
                           "0.0.0.0/0": {
```

lsheikal (changes requested):

src/genie/libs/parser/asa/tests/ShowRoute/cli/equal/golden_output_4_expected.py
thismatters: That's [exactly what I did for my first PR](https://github.com/CiscoTestAutomation/genieparser/pull/572) it is a much better solution. However: 

> all changes need to be backwards compatible

domachad: Instead of closing the original PR, you should have addressed the comments on it. This way, it is easier to keep track of comments and requests. I am asking @LukasMcClelland to help with the review since he also reviewed your first PR.

thismatters: I was interested in quickly pivoting to a different model because I have a limited amount of time to work through projects. The original PR is mentioned on the issue, which is linked above and contains many valuable insights into the ongoing work.

thismatters: @domachad Some revelations were made in the issue which exposed the appropriate path forward for handling this edge case. Unless you have strong opinions otherwise I'm going to close this PR and open another so as to keep the commit history clean.

domachad: @thismatters Ok, open a new PR, but please add @LukasMcClelland and me as reviewers

tahigash:
@thismatters as you suggested, please open new PR. I'm closing this PR.

PR #554, opened 2021-10-12
Fix ShowIsisDatabaseDetail

jmdilly:
## Description
Allow multiple adjacencies between nodes

## Motivation and Context
With the current parser, when two nodes have multiple adjacencies, only one of them is saved by the parser. This PR fixes that.

## Impact (If any)
For backward compatibility, the first key is still the neighbor ID of the adjacent node. Additionnal adjacencies have the following key have the following format '<Neighbor_ID>_#' with # >= 1.

## Screenshots:
Changes in the dict:
image

Tests :
image

## Checklist:

- [x] I have updated the changelog.
- [ ] I have updated the documentation (If applicable).
- [x] I have added tests to cover my changes (If applicable).
- [x] All new and existing tests passed.
- [ ] All new code passed compilation.

PR #586, opened 2021-12-06
Fix ShowIsisDatabaseDetail (was #554)

jmdilly:
## Description
Allow multiple adjacencies between nodes

## Motivation and Context
With the current parser, when two nodes have multiple adjacencies, only one of them is saved by the parser. This PR fixes that.

## Impact (If any)
Discussed in #554. The schema is updated with an Or so it should be backward compatible.

## Screenshots:
image

image

## Checklist:


src/genie/libs/parser/iosxe/show_isis.py
```diff
@@ -503,7 +503,14 @@ def cli(self, output=None):
                     elif mtype == 'IS':
                         is_dict = lsp_dict.setdefault('is_neighbor', {}).setdefault(ip, {})
 
-                    is_dict.update({'neighbor_id': ip,
+                    if 'metric' in is_dict.keys():
+                        if type(is_dict['metric']) is list:
+                            is_dict['metric'].append(int(group['metric']))
+                        elif type(is_dict['metric']) is int:
```
d-the-iii, line 509: this is minor, but I think it would be better to use isinstance instead of type

```suggestion
                        if isinstance(is_dict['metric'], list):
                            is_dict['metric'].append(int(group['metric']))
                        elif isinstance(is_dict['metric'], int):
```

d-the-iii (approved):
lgtm, small recommendation about type checking is all I can see that could be fixed a bit

GerriorL:
Hi @jmdilly sorry for the delay, could you please catch your fork up to the main repo so we can get this merged for you?

PR #590, opened 2021-12-09
adding support for tunneled routes

thismatters:
## Description
Changed the parser for the ASA show route command to permit multiple routes with the same subnet.

## Motivation and Context

```
S*       0.0.0.0 0.0.0.0 [1/0] via 192.2.0.1, Outside-Port-Channel
```

The output of the parser omits the first row while including the second. This is problematic because the first row represents the route that traffic would follow by default under normal conditions in the actual network. Ergo the parser output doesn't accurately reflect the config. The second line listed shows a "tunneled" route; as such it can be listed in its own section of the schema.

## Impact (If any)
A new optional key is added to the schema for `tunneled_routes` to sit alongside `routes`. When a tunneled route is found it will be added to the appropriate place.

## Checklist:
- [ ] I have updated the documentation (If applicable). N/A


thismatters:
@LukasMcClelland @domachad adding you as reviewers as requested.

src/genie/libs/parser/asa/show_route.py
```diff
@@ -51,6 +51,33 @@ class ShowRouteSchema(MetaParser):
                                 },
                             },
                         },
+                        Optional('tunneled_routes'): {
```
LukasMcClelland, line 54: Since Takashi was able to find a nice solution by separating routes into `'routes'` and `'tunneled_routes'` and since the two are identical except for name, we can now leverage a handy schema tool called `Or()`.

```
from genie.metaparser.util.schemaengine import Any, Optional, Or
...
class ShowRouteSchema(MetaParser):
    """Schema for
        * show route
    """
    schema = {
        'vrf': {
            'default': {
                'address_family': {
                    'ipv4': {
                        Optional(Or('routes', 'tunneled_routes')): {
                            Any(): {
                                'candidate_default': bool,
                                Optional('subnet'): str,
                                'route': str,
                                Optional('active'): bool,
                                Optional('date'): str,
                                Optional('route_preference'): int,
                                Optional('metric'): int,
                                Optional('source_protocol'): str,
                                Optional('source_protocol_codes'): str,
                                Optional('next_hop'): {
                                    Optional('outgoing_interface_name'): {
                                        Any(): {  # context_name for interface if there is no next_hop
                                            Optional('outgoing_interface_name'): str
                                        },
                                    },
                                    Optional('next_hop_list'): {
                                        Any(): {  # index
                                            Optional('index'): int,
                                            Optional('next_hop'): str,
                                            Optional('outgoing_interface_name'): str
                                        },
                                    },
                                },
                            },
                        },
                    },
                },
            },
        },
    }
```
This should work and will cut down on extra keys

Note: This isn't a required change since what you have currently works fine. Only implement this idea if you'd like to

src/genie/libs/parser/asa/show_route.py
```diff
@@ -126,12 +153,12 @@ def cli(self, output=None):
...
         p1 = re.sub(r'(?ms)(Codes:.+?)replicated\sroute(\s*SI\s-\sStatic\sInterVRF)?', '', out)
 
-        res = "\n".join([x.lstrip() for x in p1.splitlines()])
+        lines = [x.strip() for x in p1.splitlines()]
         entries = dict()
         last_entry = str()
         clean_lines = list()
 
-        for i, line in enumerate(res.splitlines()):
+        for line in lines:
```
LukasMcClelland, line 161: Thank you for cleaning up the code in this file 👍

LukasMcClelland (approved):

src/genie/libs/parser/asa/show_route.py
thismatters: I like the smaller change, pushing update.

domachad (approved):

PR #593, opened 2021-12-10
New Parser for Show Cable Tdr Interface

smitsialis:
## Description
Added a new parser for "show cable tdr interface {interface}" command

## Motivation and Context
Parser did not previously exist

## Impact (If any)
NIL

## Screenshots:
image

## Checklist:


src/genie/libs/parser/iosxe/show_cable_tdr_interface.py
```diff
@@ -0,0 +1,164 @@
...
+                Optional('tolerance'): int,
+                Optional('remote_pair'): str,
+                Optional('status'): str,
+            },
+            Optional('pair_d'): {
+                Optional('length'): int,
+                Optional('tolerance'): int,
+                Optional('remote_pair'): str,
+                Optional('status'): str,
+            },
```
domachad, line 53: Instead of having 4 keys with repeated values, why not have a list of pairs? In this case, the key to each pair would be its name
```suggestion
            'pairs': {
                Any() : {
                    Optional('length'): int,
                    Optional('tolerance'): int,
                    Optional('remote_pair'): str,
                    Optional('status'): str,
               }
           }            
```

domachad (changes requested):
@smitsialis Thank you for your contribution! I left a suggestion below.

src/genie/libs/parser/iosxe/show_cable_tdr_interface.py
```diff
@@ -0,0 +1,154 @@
+'''show_cable_tdr_interface.py
+IOSXE parser for the following show command
+	* show cable tdr interface {interface}
+'''
+
+# Python
+import re
+import unittest
```
domachad, line 8: please remove this import

src/genie/libs/parser/iosxe/show_cable_tdr_interface.py
```diff
@@ -0,0 +1,154 @@
...
+    Parser for:
+            show cable tdr interface {interface}
+    """
+
+    cli_command = 'show cable tdr interface {interface}'
+
+    def cli(self, interface=None, output=None):
+        if output is None:
+            if interface:
+                output = self.device.execute(self.cli_command.format(interface=interface))
```
domachad, line 51: Shouldn't interface be a mandatory field? What happens if interface is not provided?

```suggestion
    def cli(self, interface, output=None):
        if output is None:
            output = self.device.execute(self.cli_command.format(interface=interface))
```

src/genie/libs/parser/iosxe/show_cable_tdr_interface.py
```diff
@@ -0,0 +1,154 @@
...
+        else:
+            output = output
```
domachad, line 53: this is not necessary and can be removed

src/genie/libs/parser/iosxe/show_cable_tdr_interface.py
```diff
@@ -0,0 +1,154 @@
...
+                tdr_dict[interface] = {}
+                tdr_dict[interface]['interface'] = m.groupdict()['interface']
+                tdr_dict[interface]['speed'] = m.groupdict()['speed']
+
+                tdr_dict[interface]['date'] = date_performed
+                tdr_dict[interface]['time'] = time_performed
+
+                tdr_dict[interface].setdefault('pairs', {})
+
+                pair = 'pair_a'
```
domachad, line 113: can't you extract the name of the pair using regex? You have p2, p3 and p4 doing the exact same thing

domachad (changes requested):

danielgraziano (changes requested):
Same comments

src/genie/libs/parser/iosxe/show_cable_tdr_interface.py
smitsialis: @danielgraziano  I didnt want to extract it because the pair name has a space in it (ex 'Pair A').  There are only 4 pairs and thought it simpler just to do each one separately.  I guess I can extract the A B C D for pair and not include the Pair in the pair name.

smitsialis: @danielgraziano Made the changes as suggested.  Easier than I thought.
Thanks

smitsialis:
I've made all the requested changes but I guess this is now too old?  I should create a new pull request?

PR #605, opened 2022-01-20
Showsnmp contact

ahusking:
## Description
Update show snmp to capture spaces in SNMP contact

## Motivation and Context
we store information in the snmp contact field that has a space in it. with the old code we couldn't get anything past the first space

## Impact (If any)
I haven't found any

## Screenshots:
Can't provide screenshot without redacting 99% of the info

## Checklist:

- [ X] I have updated the changelog.
- [ X] I have updated the documentation (If applicable).
- [ X] I have added tests to cover my changes (If applicable).
- [ X] All new and existing tests passed.
- [ X] All new code passed compilation.


d-the-iii (approved):
Hey, thanks for the fix, we really appreciate it. before merging though, could you grab some device output that was causing the code to break and add it to the unit tests? If that isn't possible we can just merge.

ahusking:
I've updated the 2 files that run the test. essentially any config that had a space in SNMP contact caused an issue and wouldn't return anything after the space.

in my case it was "serial1_customtag serial2_customtag"

ThomasJRyan (approved):

ThomasJRyan:
Thank you for the contribution

PR #610, opened 2022-02-03
'show tacacs' parser for IOSXR

PipeFTW:
## Description
Parser for the command 'show tacacs' for IOSXR systems.

## Motivation and Context
The parser is needed in one of our projects

## Impact (If any)
-

## Screenshots:
-

## Checklist:
- [+ ] I have updated the changelog.
- [+ ] I have added tests to cover my changes (If applicable).
- [+] All new and existing tests passed.
- [+] All new code passed compilation.


src/genie/libs/parser/iosxr/show_tacacs.py
```diff
@@ -0,0 +1,146 @@
...
+            out = output
+
+        # initial dictionary
+        resultdict = dict()
+        serverdict = dict()
+
+        # RegEX for first line with vrf
+        # Server: 127.0.0.1/24 vrf=default [private]
+        p1 = re.compile(
+            r'Server\:\s(?P<ip>[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\/[0-9]+)\svrf=(?P<vrf>[A-Za-z0-9]+)\s?\[(?P<server_type>\w+)\]')
```
GerriorL, line 63: Looks like these regex statements could be improved to make them more readable/flexible if this parser needs to be expanded in the future.

For example it's best to have them start and end with ^ and $ respectively, so they don't start or end reading anywhere but the beginning/end of the line.

In addition it's also best to use things like \d, \S, or \w in lieu of [0-9] or [a-zA-Z].

src/genie/libs/parser/iosxr/show_tacacs.py
```diff
@@ -0,0 +1,146 @@
...
+    def cli(self, output=None):
+        if output is None:
+            cmd = self.cli_command[0]
+            out = self.device.execute(cmd)
+        else:
+            out = output
+
+        # initial dictionary
+        resultdict = dict()
+        serverdict = dict()
```
GerriorL, line 58: Using serverdict as the main dictionary, and resultdict as a subdictionary might be confusing; typically the main dictionary that gets returned is named ret_dict, but it's not a mandatory change in my opinion.

GerriorL, line 58: Typically it's preferred to use literal declaration when possible, {} instead of dict(), but again I think that's an optional change.

GerriorL (changes requested):
Hi there! Thank you for your submission. It looks great, but there are a couple of small changes I'd like to request.

GerriorL (approved):
Great job on this parser! In the future you don't need to bother with the tests/show_tacacs.py file, just the folder based unittest is sufficient.

PipeFTW:
> Great job on this parser! In the future you don't need to bother with the tests/show_tacacs.py file, just the folder based unittest is sufficient.

Oh ok, I have another pull request open, have there a testfile too. Need to do 3-4 other command parser as well. Thanks for the info :)

lsheikal (approved):

PR #614, opened 2022-02-07
Add 3.10 to workflow

GerriorL:
## Description
Adds python 3.10 to workflow

## Motivation and Context
External contributions need to be tested on 3.10 now that pyATS/Genie supports that version.

## Checklist:

### DIFF
--- a/.github/workflows/run_tests.yml
+++ b/.github/workflows/run_tests.yml
@@ -14,7 +14,7 @@ jobs:
 
     strategy:
       matrix:
-        python-version: [3.6, 3.7, 3.8, 3.9, '3.10']
+        python-version: ['3.6', '3.7', '3.8', '3.9', '3.10']
 
     steps:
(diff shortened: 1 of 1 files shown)


.github/workflows/run_tests.yml
```diff
@@ -14,7 +14,7 @@ jobs:
 
     strategy:
       matrix:
-        python-version: [3.6, 3.7, 3.8, 3.9]
+        python-version: [3.6, 3.7, 3.8, 3.9, '3.10']
```
tahigash, line 17: interesting to know needing `'`. maybe better to have on all of them to be consistent.

tahigash (approved):
approved with comment

.github/workflows/run_tests.yml
tahigash: confirmed the fix.

Issue #619, opened 2022-02-15
ASA parsing of show inventory fails 


empusas:
Parsing of the "show Inventory" command fails on one of our ASA firewalls:

From the sensible script:

`FAILED! => {"msg": "parse_genie: Missing keys: [['Ethernet 1/13', 'pid'], ['Ethernet 1/13', 'vid'], ['Ethernet 1/13', 'sn'], ['Ethernet 1/14', 'pid'], ['Ethernet 1/14', 'vid'], ['Ethernet 1/14', 'sn']] - Failed to parse command output."}
`
from the command output:

```
Name: "Ethernet 1/13", DESCR: "10gbase-lr"
PID: RTXM228-401-C98, VID: , SN: ACW111115H9

Name: "Ethernet 1/14", DESCR: "1glx"
PID: FTLF1318P3BTL-C, VID: , SN: FNS11111J1K
```

lsheikal:
Thanks for bringing up the issue. Looks like the parser couldn't accommodate this console output, You might want to contribute for the parser. For it, there is the reference how to develop parser. And there are very nice YouTube video in the page by pyATS users.

https://pubhub.devnetcloud.com/media/pyats-development-guide/docs/writeparser/writeparser.html

muhammad-rafi:
Sorry for hijacking the thread, I was looking at one of my other pull request and get here, I think I can fix this one but, just one question, is ASA sending empty output like this 
VID: ,

Not like this ? 
VID: N/A

empusas:
Hi,

I pasted exactly the output of the command line, so "VID: ," is correct.

muhammad-rafi:
Thanks for the response, can you please also mention the asa os version you are using?, so I can test against that version and other versions to see it if this doesn't break for others

lsheikal:
@muhammad-rafi kindly link PR to this issue for reference.

muhammad-rafi:
> @muhammad-rafi kindly link PR to this issue for reference.

I have tested this in the lab and it works fine but I have a pull request currently opened #615 and been 17 days and waiting for someone to run the checks and review, once that done, I will then add this one and raise another pull request to fix this, or if like me to include this one on the same pull request? I can do this too, I will update the change log. Let me know what would you prefer

lsheikal:
Kindly create a new PR for this fix.

muhammad-rafi:
> Kindly create a new PR for this fix.

Pull requested created, please review #630

PR #630, opened 2022-03-01
ShowInventory ASA fix#619

muhammad-rafi:
## Description
Updated the regex to capture empty value of VID in 'show inventory' output mentioned in the issue #619 

## Motivation and Context
Resolve the empty value of VID in 'show inventory' output for ASA

## Impact (If any)
None 

## Checklist:


src/genie/libs/parser/asa/show_inventory.py
```diff
@@ -49,14 +49,13 @@ def cli(self, output=None):
 
         # Name: "Chassis", DESCR: "ASA 5555-X with SW, 8 GE Data, 1 GE Mgmt"
         # Name: "power supply 1", DESCR: "ASA 5545-X/5555-X AC Power Supply"
-        p1 = re.compile(
-            r'^Name: +"+(?P<name>.+)"+,* +DESCR:+ "+(?P<description>.+)+"$')
+        p1 = re.compile(r'^Name:\s+"+(?P<name>.+)",\s+DESCR:\s"+(?P<description>.+)"$')
```
d-the-iii, line 52:
```suggestion
        p1 = re.compile(r'^Name:\s+"(?P<name>.+)",\s+DESCR:\s+"(?P<description>.+)"$')
```

src/genie/libs/parser/asa/show_inventory.py
```diff
@@ -49,14 +49,13 @@ def cli(self, output=None):
...
 
         # PID: ASA5555, VID: V01, SN: AAAAA11111
         # PID: AAA-AAA-AAA, VID: N/A, SN: AAA111
         # PID: N/A, VID: N/A, SN: AAAAA11111
-        p2 = re.compile(r'^PID: +(?P<pid>.+)( )?,+ VID: (?P<vid>.+)( )?, '
-            '+SN: (?P<sn>.+)$')
+        # PID: N/A, VID: , SN: AAAAA11111
+        p2 = re.compile(r'^PID:\s+(?P<pid>.+),+\sVID:(?P<vid>.+),\s+SN:(?P<sn>.+)$')
```
d-the-iii, line 58: This captures the space with the keys, which you remove in the for loop. It is better capture correctly the first time. I think it's fine to leave the space replacement in, it will matter for when only a space is captured, but I have altered the regex to not capture the extra spaces.

```suggestion
        p2 = re.compile(r'^PID:\s+(?P<pid>.+),\s+VID:\s?(?P<vid>.+),\s+SN:\s+(?P<sn>.+)$')
```

src/genie/libs/parser/asa/tests/ShowInventory/cli/equal/golden_output_expected.py
```diff
@@ -1,26 +1,26 @@
 expected_output = {
-    "Chassis": {
-        "description": "ASA 5555-X with SW, 8 GE Data, 1 GE Mgmt",
-        "pid": "ASA5555",
-        "vid": "V01",
-        "sn": "AAAAA11111",
+    'Chassis': {
```
d-the-iii, line 2: please don't switch the quoting of the whole unit test

src/genie/libs/parser/asa/tests/ShowInventory/cli/equal/golden_output_output.txt
```diff
@@ -1,5 +1,4 @@
-
-ciscoasa> show inventory
+ciscoasa# show inventory
```
d-the-iii, line 1: I don't know why this was altered but the whole line should actually be removed so that there is only device output.

d-the-iii (changes requested):

src/genie/libs/parser/asa/show_inventory.py
muhammad-rafi: looks like my typo, thanks for pointing that out

src/genie/libs/parser/asa/show_inventory.py
muhammad-rafi: I dont know what you mean by captures the spaces with the keys, the one I removed the one which was not capturing the empty output for VID: , I don't see much difference between your regex and mine, I tested on both now and had same results.

src/genie/libs/parser/asa/tests/ShowInventory/cli/equal/golden_output_expected.py
muhammad-rafi: The reason I did because I normally keep single quote to differentiate between the json object or python dict, but it doesn't really matter here. Happy with whatever you go with :)

src/genie/libs/parser/asa/tests/ShowInventory/cli/equal/golden_output_output.txt
muhammad-rafi: Agree on this one, line should be removed.

muhammad-rafi (commented):
Updated as suggested, thank you for the advices

src/genie/libs/parser/asa/show_inventory.py
```diff
@@ -80,9 +79,11 @@ def cli(self, output=None):
...
                 vid = vid.replace(' ','')
+                sn = groups['sn']
+                sn = sn.replace(' ','')
                 dict_name.update({'pid': pid})
                 dict_name.update({'vid': vid})
-                dict_name.update({'sn': groups['sn']})
+                dict_name.update({'sn': sn})
                 continue
 
         return ret_dict
```
lsheikal, line 89: new line needed.

src/genie/libs/parser/asa/tests/ShowInventory/cli/equal/golden_output_output.txt
```diff
@@ -1,5 +1,3 @@
-
-ciscoasa> show inventory
```
lsheikal, line 2: why these lines been removed from the Unittest?

lsheikal (changes requested):

src/genie/libs/parser/asa/tests/ShowInventory/cli/equal/golden_output_output.txt
tahigash: this is ok. we can have only show command output excluding command itself.

d-the-iii: I requested that, it is a command line prompt and not device output, it's better to not include that.

muhammad-rafi (commented):
updated as requested

src/genie/libs/parser/asa/show_inventory.py
muhammad-rafi: added new line.

lsheikal (approved):

d-the-iii (approved):
LGTM, approved.

Just to explain the regex change for your, here are two screenshots from regex101, the first is your original regex, the second is my slightly altered one.

Screen Shot 2022-03-03 at 10 16 37 AM

Screen Shot 2022-03-03 at 10 16 48 AM

It's subtle, but in the first one you can it captures ' V01' and ' AAAAA11111', whereas the second one captures 'V01' and 'AAAAA11111' without a leading space.

muhammad-rafi:
> 

This is from my expression 
(py_3.7) murafi@MURAFI-M-VC10:my_parsers$ python show_inventory_asa.py 
{
    'Ethernet 1/13': {'description': '10gbase-lr', 'pid': 'RTXM228-401-C98', 'vid': '', 'sn': 'ACW111115H9'},
    'Ethernet 1/14': {'description': '1glx', 'pid': 'FTLF1318P3BTL-C', 'vid': '', 'sn': 'FNS11111J1K'}
}

This is from your expression 
(py_3.7) murafi@MURAFI-M-VC10:my_parsers$ python show_inventory_asa.py 
{
    'Ethernet 1/13': {'description': '10gbase-lr', 'pid': 'RTXM228-401-C98', 'vid': '', 'sn': 'ACW111115H9'},
    'Ethernet 1/14': {'description': '1glx', 'pid': 'FTLF1318P3BTL-C', 'vid': '', 'sn': 'FNS11111J1K'}
}
(py_3.7) murafi@MURAFI-M-VC10:my_parsers$ 

Let me know if you spot any difference up there. 

Take a look in the code for below lines, I am removing the white space anyways, so no difference between your and mine expression in this case, I originally did this because previous author create expression was not capturing the empty output, so I captured leading white space on purpose. 

vid = groups['vid']
vid = vid.replace(' ','')
sn = groups['sn']
sn = sn.replace(' ','')

Obviously there are multiple ways to solve these, therefore, I dont mind either :)

PR #658, opened 2022-05-10
New parser for 'show vlan brief' on Cisco IOSXE

barfa1:
## Screenshots:

Screenshot (994)
Screenshot (995)


GerriorL (changes requested):
Hi Shubham,
This looks good, but I think there's a better approach you could take in the schema.

src/genie/libs/parser/iosxe/show_vlan_brief.py
```diff
@@ -0,0 +1,98 @@
+"""
```
omehrabi, line 1: we already have show_vlan for iosxe please add the parser to that file instead of creating a new one

omehrabi (changes requested):
Hi,
thank you for contributing left comment

PR #667, opened 2022-06-06
Updated source_template parser

obrigg:
## Description
A bug is simply a situation we haven't thought of.
While parsing a "show run" for an interface, I hit the following line `source template 802.1X_PORT_AUTH_TEMPLATE` that did not parse the source template attribute.
The current parser doesn't handle special characters in the templates' names.

## Motivation and Context
Allow the parser to handle special characters in the templates' names.

## Checklist:

### DIFF
--- a/src/genie/libs/parser/iosxe/show_run.py
+++ b/src/genie/libs/parser/iosxe/show_run.py
@@ -646,8 +646,8 @@ class ShowRunInterface(ShowRunInterfaceSchema):
         # switchport port-security mac-address sticky 1020.4bb1.6f2f
         p61 = re.compile(r"^switchport port-security mac-address sticky (?P<value>([a-fA-F\d]{4}\.){2}[a-fA-F\d]{4})$")
 
-        # source template USER_NoAuth
-        p62 = re.compile(r"^source template (?P<template>\w+)$")
+        # source template USER_No_802.1X_Auth
+        p62 = re.compile(r"^source template (?P<template>[a-zA-Z0-9@#$&()-`.+,/]*)$")
 
         # host-reachability protocol bgp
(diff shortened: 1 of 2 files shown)


src/genie/libs/parser/iosxe/show_run.py
```diff
@@ -646,8 +646,8 @@ def cli(self, interface=None, output=None):
         # switchport port-security mac-address sticky 1020.4bb1.6f2f
         p61 = re.compile(r"^switchport port-security mac-address sticky (?P<value>([a-fA-F\d]{4}\.){2}[a-fA-F\d]{4})$")
 
-        # source template USER_NoAuth
-        p62 = re.compile(r"^source template (?P<template>\w+)$")
+        # source template USER_No_802.1X_Auth
+        p62 = re.compile(r"^source template (?P<template>[a-zA-Z0-9@#$&()-`.+,/]*)$")
```
domachad, line 650: Can you add more examples where all these special characters are needed?
A unit test that covers this change would be needed as well

Moreover, I think this pattern can be simplified. I believe that this could achieve the same results"
```suggestion
        # source template USER_NoAuth
        # source template USER_No_802.1X_Auth
        p62 = re.compile(r"^source template (?P<template>[\w#$&-`.]+)$")
```

domachad (changes requested):
@obrigg Thank you for contributing to pyATS. Please see the comment below regarding your change in `p62`

Taarini (changes requested):
Can you please add changelog

obrigg:
Anything else?

domachad:
> Anything else?

You haven't provided a folder based unit test that covers this change

domachad (changes requested):
Add a Unit Test

Issue #668, opened 2022-06-06
When there is no eigrp neighbor, parsed output is not empty


balmasea:
I'm parsing the eigrp neighbor for a bunch of defines and when there are no neighbors, output is different than expected.
I would expect to receive an empty yml like this
```yaml
''
```

as I'm getting for `show ip ospf neighbor`. For `show ip eigrp neighbor` I'm getting this output though:
```yaml
EIGRP-IPv4 Neighbors for AS(XXXX)
...
```

balmasea:
FYI, I'm already working on the code to fix this issue.

balmasea:
I was wrong, it was a problem in our code.

PR #669, opened 2022-06-06
Fixing #668. checking that there are eigrp neighbors; otherwise, returning empty dict

balmasea:
## Description
Returning empty dictionary when there is not any eigrp neighbor.

## Motivation and Context
It allows to keeps parsing when device has no eigrp neighbor.

## Impact (If any)
N/A

## Screenshots:
N/A
## Checklist:


balmasea:
Just realized it is not necessary. Apologies.

PR #701, opened 2022-09-05
Added AOS Show Commands

Alexpf20210007:
## Description
* AOS
    * Added ShowInterfacesBrief:
        * show interfaces brief
    * Added ShowVlan:	
        * show vlan
    * Added ShowInterfacesTransceiver:
        * show interfaces transceiver
    * Added ShowStacking:
        * show stacking

## Motivation and Context
The commands for Aruba OS are different than Cisco IOS, IOSXE, IOSXR, etc.
## Checklist:

- [X ] I have updated the changelog.
- [X ] I have added tests to cover my changes (If applicable).
- [ ] All new and existing tests passed. - I have not tested the test files.


Alexpf20210007:
I recently got Aruba OS (aos) merged into the unicon plugin. This opens Aruba OS to anybody that wants to contribute to the genieparser project. I have already created 4 show commands to get started with.

Alexpf20210007:
I have an SDK folder with sample output results. Please let me know if I need to remove that.

src/genie/libs/parser/aos/show_vlan.py
```diff
@@ -0,0 +1,108 @@
...
+    """Parser for show vlan"""
+    cli_command = 'show vlan'
+
+    def cli(self, output=None):
+        if output is None:
+            out = self.device.execute(self.cli_command)
+        else:
+            out = output
+
+#   This matches the report portion of the show vlan command.
```
GerriorL, line 55: I think that the file could be cleaned up a bit by having the comments at the appropriate level of indentation for the code

src/genie/libs/parser/aos/show_vlan.py
```diff
@@ -0,0 +1,108 @@
...
+class ShowVlanSchema(MetaParser):
+    """Schema for show vlan"""
+    schema = {
+        'vlans':{
+            Any():{
+                Optional('vlan_id'): str,
+                Optional('name'): str,
+                Optional('status'): str,
+                Optional('voice'): str,
+                Optional('jumbo'): str,
```
GerriorL, line 36: Do all of these need to be optional? Optional keys are best used when it's possible for only part of a line to show up. Overusing optional can lead to parsers passing when they should otherwise have failed

GerriorL (changes requested):
Hi Alex! Thank you for this submission, and for the initiative in getting AOS supported. I do have some feedback.

Firstly, there is no need to create a new file for each parser. Parsers that operate off the same or closely related parent commands can be added to the same file. For examples all of the 'show interfaces' commands could be added to a single 'show_interfaces.py'; there is no need to have interfaces brief and interfaces transceiver listed separately. 

Next, when it comes to writing the tests for your parser I'd like to refer you to our contribution guide, where you can find instructions for creating a [Folder Test](https://pubhub.devnetcloud.com/media/pyats-development-guide/docs/writeparser/writeparser.html#folder-based-testing).

When it comes to the parsers themselves, I should mentions that schema keys should only be made optional only when there's no other way to address the device's possible outputs. For the commands you're parsing here, if there is a column in the output table that may or may not be filled, then an optional key is helpful for that column only. 

Lastly it would be best to remove the sdk/ folder as you mentioned.

PR #726, opened 2023-01-24
fixed an issue where show cdp neighbors detail didnt parse the vtp-do…

hoferchr:
…main correctly.

## Description
vtp-domains can contain not only characters, but also numbers. I expanded the regex for numbers 0-9.

## Motivation and Context
We have vtp-domains which contains characters and numbers e.g. 'F16' or 'Z990'

## Impact (If any)
none

### DIFF
--- a/src/genie/libs/parser/iosxe/show_cdp.py
+++ b/src/genie/libs/parser/iosxe/show_cdp.py
@@ -248,7 +248,7 @@ class ShowCdpNeighborsDetail(ShowCdpNeighborsDetailSchema):
         # VTP Management Domain: 'Accounting Group'
         vtp_management_domain_re = re.compile(r'^VTP\s*Management\s*'
                                               r'Domain\s*:\s*'
-                                              r'\W*(?P<vtp_management_domain>([a-zA-Z\s]+'
+                                              r'\W*(?P<vtp_management_domain>([a-zA-Z0-9\s]+'
                                               r'))\W*')
 
         # Holdtime : 126 sec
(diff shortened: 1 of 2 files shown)


Taarini (approved):

src/genie/libs/parser/iosxe/show_cdp.py
```diff
@@ -248,7 +248,7 @@ def cli(self, output=None):
         # VTP Management Domain: 'Accounting Group'
```
domachad, line 248: can you please an example of a string that would support this change?

domachad (changes requested):
Please add one or more unit tests that support your change

Taarini:
Hi

Can you please address the review comments.

hoferchr:
Hi domachad,
we use numerical VTP Management Domains. For example:

VTP Management Domain: '4074'

If you need anything more, please let me know.
thanks,
Christian

src/genie/libs/parser/iosxe/show_cdp.py
hoferchr: VTP Management Domain: '4074'

domachad: I am sorry, I meant "please add an example". Just add a string on top of the existing one

hoferchr: Hi domachad, I added Test-examples. Please let me know what else is needed. I am sorry but this is my first pull-request, so I kindly ask for your support here. thank you!

PR #752, opened 2023-05-02
Added parsers for ShowInterfacesCapabilities

smitsialis:
## Description
Add new parsers for Show Interfaces Capabilities

## Motivation and Context
Not previously there and need to use it in my work

## Impact (If any)
No impact

## Screenshots:
image

## Checklist:

- [x ] I have updated the changelog.
- [x ] I have added tests to cover my changes (If applicable).
- [x ] All new and existing tests passed.
- [x ] All new code passed compilation.


smitsialis:
@GerriorL, @lsheikal, @danielgraziano   
Hi, I've had this MR here for 3 wks and not seen any reviews.  I've tried to add this code previous but had to do a new MR each time because the MR became too old to merge.  Is there something different I should be doing?
Thanks,

PR #779, opened 2023-08-15
Hvrp display configuration dhcp

nielsvanhooy:
## Description
added new command in hvrp
*DisplayConfigDhcp

## Motivation and Context
i needed schema's to extract configuration settings for dhcp

## Impact (If any)
no impact.

## Checklist:


src/genie/libs/parser/hvrp/display_config_dhcp.py
```diff
@@ -0,0 +1,182 @@
...
+            Optional("dhcp_options"): {
+                Any(): {
+                    Optional("option"): str,
+                    Optional("type"): str,
+                    Optional("data"): str,
+                }
+            },
+            Optional("lease_time"): str,
+        },
+    }
```
GerriorL, line 35: Schemas are used to validate the key and value type of the output to ensure it's consistent. For that reason the topmost level key should never be an Any(). It's also best to not have keys set as Optional() unnecessarily. It's best to make every key possible be mandatory.

GerriorL (changes requested):
Hi! Thanks for the contribution.

I understand what you're trying to accomplish with this contribution but unfortunately this implementation goes against the project's code standards. What I'd recommend is to create a full parser for `display current-configuration`. Then you can create a genielibs get API, and from that extract specifically the information related to the dhcp config by calling the full parser and only return the desired keys.

PR #804, opened 2023-12-10
Standardize VRF Regex for IOSXE show_vrf.py

mattchilders:
Currently ShowVrf doesn't match all valid special characters in VRF names, and uses different regexes for different commands throughout show_vrf.py. This change standardizes the regexes to all be the same and matches the special characters.

## Description
This is a minor commit that standardizes the vrf regex in "show_vrf.py" so that they all use the [\S\]+ rather than the mix of regexes that matched only certain special characters. This makes iosxe's show_vrf.py comparable with nxos's show_vrf.py. 

## Motivation and Context
IOS-XE parser didn't match VRF's with special characters.

## Impact (If any)
None that I'm aware of

## Screenshots:
Screenshot's of the relevant tests that had classes changed.
ShowVrf
image
ShowVrfDetail
image
ShowVRFIPv6
image
ShowPlatformSoftwareCefIpVrf
image

## Checklist:

- [ x ] I have updated the changelog.
- [ x ] I have added tests to cover my changes (If applicable).
- [ x ] All new and existing tests passed.

### DIFF
--- a/changelog/undistributed/change_log_show_vrf_iosxe_20231209154701.rst
+++ b/changelog/undistributed/change_log_show_vrf_iosxe_20231209154701.rst
@@ -1,0 +1,6 @@
+--------------------------------------------------------------------------------
+                            Fix
+--------------------------------------------------------------------------------
+* IOSXE
+    * Modified ShowVrf:
+        * Standardized VRF Matching to include special characters beyond dash and dot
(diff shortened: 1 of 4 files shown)


ThomasJRyan (approved):
Thank you for the contribution! I like the unit test you added. Never really considered having more generic tests like that

PR #843, opened 2024-04-26
Fix regex and both for import and export route-targets NXOS

BingPB:
## Description

I have changed the regex pattern for matching route-target to include ':'.
A ':' is required when configuring a import/export route-target.

I have also fixed, so a route-target can be 'both' imported and exported at the same time.

## Motivation and Context

This fixes the "show running-config vrf {vrf} | sec '^vrf'" command on NXOS.
Before route-targets would not be parsed, and not appear in parsed output.

## Impact (If any)

No negative impact, as this did not work previously.

## Checklist:

### DIFF
--- a/src/genie/libs/parser/nxos/show_vrf.py
+++ b/src/genie/libs/parser/nxos/show_vrf.py
@@ -292,7 +292,7 @@ class ShowRunningConfigVrf(ShowRunningConfigVrfSchema):
         #     route-target both auto
         #     route-target both auto mvpn
         #     route-target both auto evpn
-        p5 = re.compile(r'^\s*route-target +(?P<rt_type>\w+) +(?P<rt>\w+)( +(?P<rt_evpn_mvpn>\w+))?$')
+        p5 = re.compile(r'^\s*route-target +(?P<rt_type>\w+) +(?P<rt>\w+:\w+)( +(?P<rt_evpn_mvpn>\w+))?$')
 
         # find all list of vrfs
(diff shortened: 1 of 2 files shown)


BingPB:
Hello, any update on this? Would be nice to have the library fixed.
@Taarini @lsheikal

src/genie/libs/parser/nxos/show_vrf.py
```diff
@@ -292,7 +292,7 @@ def cli(self, vrf=None):
         #     route-target both auto
         #     route-target both auto mvpn
         #     route-target both auto evpn
-        p5 = re.compile(r'^\s*route-target +(?P<rt_type>\w+) +(?P<rt>\w+)( +(?P<rt_evpn_mvpn>\w+))?$')
+        p5 = re.compile(r'^\s*route-target +(?P<rt_type>\w+) +(?P<rt>\w+:\w+)( +(?P<rt_evpn_mvpn>\w+))?$')
```
Taarini, line 295: Please add comment for the regex change

Taarini (changes requested):
Thanks for your contribution. Sorry for not reviewing it sooner

Could you also add UT

src/genie/libs/parser/nxos/show_vrf.py
BingPB: Sorry, but I'm not sure what to add since it should have been like this from the start.

PR #923, opened 2025-01-27
[NXOS] Modified ShowInterfaceStatus parser to accommodate the value 'adminCfgC'

srbharadwaj:
## Description
Enhance the existing ShowInterfaceStatus parser to accommodate the value 'adminCfgC' for status field 

## Motivation and Context
Status field is hard coded, one the values needed for ZR/ZRP related testcases was 'adminCfgC'


ThomasJRyan (changes requested):
The change seems fine, if you could add a unittest for the new output that would be great!

PR #930, opened 2025-02-23
"show authentication sessions" returns  "Parser Output is empty" when "No sessions currently exist" cli output is returned from device.

czinu:
## Description

When there are no authentication sessions, this is when CLI outputs:

```
access-switch-01#show authentication sessions
No sessions currently exist
```

 `` Unable to parse output for command 'show authentication sesssions" (Parser Output is empty)`` error is returned.

## Motivation and Context
Return proper response when authentication service is not enabled or there are currently no active sessions. 

```
{
    "interfaces": {

    },
    "session_count": 0
}
```

Ansible failed run:
```
fatal: [access-switch-01]: FAILED! =>
  msg: Unable to parse output for command 'show authentication sessions' (Parser Output is empty)
```

Ansible successful run:
```
ok: [access-switch-01] =>
  msg:
    show_authentication_sessions:
      interfaces: {}
      session_count: 0
```

### DIFF
--- a/src/genie/libs/parser/iosxe/show_authentication_sessions.py
+++ b/src/genie/libs/parser/iosxe/show_authentication_sessions.py
@@ -82,6 +82,9 @@ class ShowAuthenticationSessions(ShowAuthenticationSessionsSchema):
         else:
             out = output
 
+        if "No sessions currently exist" in out:
+            return {"interfaces": {}, "session_count": 0}
+
         # initial return dictionary
         ret_dict = {}
 
--- a/changelog/undistributed/fix_show_authentication_sessions.rst
+++ b/changelog/undistributed/fix_show_authentication_sessions.rst
@@ -1,0 +1,6 @@
+--------------------------------------------------------------------------------
+                            Fix
+--------------------------------------------------------------------------------
+* IOSXE
+    * Modified ShowAuthenticationSessions:
+        * Added default value for "No sessions currently exist" cli output.
--- a/src/genie/libs/parser/iosxe/tests/ShowAuthenticationSessions/cli/equal/golden_output_4_expected.py
+++ b/src/genie/libs/parser/iosxe/tests/ShowAuthenticationSessions/cli/equal/golden_output_4_expected.py
@@ -1,0 +1,6 @@
+expected_output = {
+    "interfaces": {
+
+    },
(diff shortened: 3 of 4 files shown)


src/genie/libs/parser/iosxe/show_authentication_sessions.py
```diff
@@ -82,6 +82,9 @@ def cli(self, interface=None, output=None):
         else:
             out = output
 
+        if "No sessions currently exist" in out:
+            return {"interfaces": {}, "session_count": 0}
```
ThomasJRyan, line 86: Typically, if there's no output to parse we would rather return an empty dictionary and allow an `EmptySchemaError` to be raised

ThomasJRyan (changes requested):
While we do appreciate the contribution, it does kind of go against how we prefer to structure our parsers. Is there perhaps a middle ground we can reach?

src/genie/libs/parser/iosxe/show_authentication_sessions.py
czinu: To be honest in my opinion there is a data to parse here, because there is output being returned. While I agree that Cisco could do it better and perhaps output just headers which would result in the empty dict, but it was their decision to output "No sessions currently exist". I would agree on empty dict being returned but why do you think that should raise any kind of error? This feature cannot be technically "disabled".

ThomasJRyan: It's simply how Genie works. Empty dictionaries raise errors to allow the developer to know that nothing meaningful was parsed from the CLI output, and allows them to handle that accordingly. It's extremely useful when you're calling a parser five levels down a stack and want to quickly propagate that up without requiring several layers of checks.

I can get behind the idea that "No sessions currently exist" qualifies as meaningful data however. With that in mind, I'll need a changelog added [here](https://github.com/CiscoTestAutomation/genieparser/tree/main/changelog/undistributed) before I can approve this PR

czinu: I haved added relared changelog file. Thanks for this quick review.

ThomasJRyan (approved):

Taarini (approved):

PR #964, opened 2025-10-24
Handle NTP servers configured to use a key

jlmcgraw:
## Description
Handle NTP servers configured to use a key

eg:
ntp server vrf Mgmt 10.2.2.2 key 2
ntp server vrf Mgmt 10.3.3.3 key 3 prefer

## Motivation and Context
NTP servers that have a key configured are not properly parsed by the current regex

## Impact (If any)
NTP servers that have a key configured are properly parsed 

## Checklist:


ThomasJRyan (changes requested):
One small change to the changelog and all is good to go

ThomasJRyan (approved):

PR #1003, opened 2026-09-11
feat(nxos): add show ip ospf neighbors parser

behrang-t:
NX-OS `show ip ospf neighbors` summary output needs a dedicated parser for structured neighbor data.

This PR adds `ShowIpOspfNeighbors` with its schema, parser registry entry, fixtures, regression tests and changelog.

The parser extracts the process, VRF, total neighbor count, interface, neighbor router ID, priority, state, role, uptime and address. Neighbors are grouped by interface so the same router ID on separate links is preserved.

An explicit zero-neighbor summary produces structured output. Completely empty input raises `SchemaEmptyParserError`. Missing counts, duplicate neighbors on the same interface and mismatches between the declared count and parsed rows raise errors rather than returning incomplete results.

Validation:

* All 9 focused regression tests passed, including command-based parser lookup.
* The official folder-based runner passed all 5 golden fixtures and 1 empty fixture.
* Successfully parsed 20 recorded NX-OS outputs from 7 lab devices.
* `git diff --cached --check` passed.

`make json` was executed. The committed registry diff retains only the new OSPF command entry, excluding unrelated generator changes.

### DIFF
--- a/src/genie/libs/parser/nxos/show_ospf.py
+++ b/src/genie/libs/parser/nxos/show_ospf.py
@@ -12,6 +12,7 @@ NXOS parsers for the following show commands:
     * show ip ospf interface
     * show ip ospf interface vrf <WORD>
     * show ip ospf neighbors detail
+    * show ip ospf neighbors
     * show ip ospf neighbors detail vrf <WORD>
     * show ip ospf database external detail
     * show ip ospf database external detail vrf <WORD>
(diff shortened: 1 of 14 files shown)


src/genie/libs/parser/nxos/show_ospf.py
```diff
@@ -2352,6 +2353,114 @@ def cli(self, vrf='', interface='', output=None):
         return ret_dict
 
 
+# ======================================================
+# Schema for 'show ip ospf neighbors'
+# ======================================================
+class ShowIpOspfNeighborsSchema(MetaParser):
+    """Schema for the NX-OS OSPF neighbor summary."""
```
Sripadvallabh: Please make the schema docstring list the exact supported command (`show ip ospf neighbors`). Parser schema and parser class docstrings should identify their associated CLI rather than only describing the output generally.

src/genie/libs/parser/nxos/show_ospf.py
```diff
@@ -2352,6 +2353,114 @@ def cli(self, vrf='', interface='', output=None):
...
+
+        for line in output.splitlines():
+            line = line.strip()
+            if not line:
+                continue
+            match = p_header.match(line)
+            if match:
+                group = match.groupdict()
+                instance = result.setdefault('vrf', {}).setdefault(
+                    group['vrf'], {}).setdefault('address_family', {}).setdefault(
```
Sripadvallabh: Please wrap the newly added non-regex source and test lines at 80 characters. This chain and several exception/test lines below exceed that limit; regex patterns and changelog content are exempt.

Sripadvallabh (changes requested):
Fix comments & click re-review

src/genie/libs/parser/nxos/show_ospf.py
behrang-t: Updated the schema docstring to identify the exact supported command: `show ip ospf neighbors`.

src/genie/libs/parser/nxos/show_ospf.py
behrang-t: Wrapped all newly added non-regex source and test lines to 80 characters, leaving regex and changelog content unchanged as requested. The 9 focused regression tests and the folder-based tests (5 golden and 1 empty fixture) pass.

Sripadvallabh (approved):

